PR #26, opened 2017-07-20
Add dedicated cloning of repositories

fineguy:


vmarkovtsev:
This is an easy to test functionality - please add tests.

vmarkovtsev:
Also, please mind the format of commit messages:

[Capital letter]-verb ... [no dot in the end]

In your case

Add the cloning command

ast2vec/clone_repo.py
```diff
@@ -0,0 +1,144 @@
...
+from multiprocessing.dummy import Pool
+import os
+import shutil
+import subprocess as sb
+import zlib
+
+from ast2vec.repo2.base import LinguistFailedError
+
+# Set up logging
+logging.basicConfig(
```
vmarkovtsev: Nope, this is already done in `__main__.main()`, please remove

ast2vec/clone_repo.py
```diff
@@ -0,0 +1,144 @@
...
+    format='%(asctime)s - %(name)s - %(levelname)s - %(thread)d - %(message)s')
+LOGGER = logging.getLogger("clone_repo")
```
vmarkovtsev: Please avoid this global var and init the logger inside the functions

ast2vec/clone_repo.py
```diff
@@ -0,0 +1,144 @@
...
+            list(generator_repo_urls(inputs)))
+
+
+def clone_repositories(args):
+    LOGGER.setLevel(args.log_level)
+    clone_args = _sanitize_kwargs(args)
+    clone_repos(**clone_args)
+
+
+def generator_repo_urls(inputs):
```
vmarkovtsev: `generate_repo_urls`

ast2vec/tests/test_repo2coocc.py
```diff
@@ -45,10 +45,11 @@ def test_asdf(self):
             validate_asdf_file(self, file.name)
 
     def test_linguist(self):
+        # Make sure provide path to linguist is not pointing to an executable
         with self.assertRaises(FileNotFoundError):
             Repo2Coocc(linguist="xxx", timeout=600)
         with self.assertRaises(FileNotFoundError):
-            Repo2Coocc(linguist=__file__, timeout=600)
+            Repo2Coocc(linguist=os.path.dirname(__file__), timeout=600)
```
vmarkovtsev: It must point to the executable. Something is missing or wrong

ast2vec/__main__.py
```diff
@@ -6,6 +6,7 @@
 from ast2vec.vw_dataset import nbow2vw_entry
 from modelforge.logs import setup_logging
 
+from ast2vec.clone_repo import clone_repositories
```
vmarkovtsev: Let's name it `cloning`

ast2vec/tests/test_repo2coocc.py
fineguy: The point of this test is to check for missing enry. However, repo2.base.py raises FileNotFoundError only when the provided path is not an executable (otherwise shutil.which will point to the executable).

ast2vec/cloning.py
```diff
@@ -0,0 +1,178 @@
...
+
+        try:
+            sb.check_output(["git", "clone", "--depth=1", git_url, repo_dir],
+                            env=env, stdin=sb.DEVNULL, stderr=sb.STDOUT)
+        except sb.CalledProcessError as e:
+            shutil.rmtree(repo_dir, ignore_errors=True)
+            self._log.error("Git failed to clone repo. git stderr:\n\t" +
+                            "\n\t".join(e.output.decode("utf8").split("\n")))
+            if not ignore:
+                raise e from None
```
vmarkovtsev: Better to leave Konst's code, shit happens
https://github.com/src-d/ast2vec/pull/26/commits/b86ec786674c34bbc07a9f3abf53c2b94dc2da7c#diff-7259728938d15cd53393199ce4d87effL92

ast2vec/cloning.py
```diff
@@ -0,0 +1,178 @@
...
+        self._linguist, self._is_enry = self._find_enry(linguist)
+        self._num_threads = num_threads
+
+    def clone_repo(self, git_url, ignore, target_dir, redownload=False):
+        git_url = self._prepare_repo_url(git_url)
+
+        try:
+            repo_dir = self._prepare_repo_dir(git_url, target_dir, redownload)
+        except OSError as e:
+            if e.errno == errno.EEXIST:
```
vmarkovtsev: `except FileNotFound`. This exception does not exist in py2, but we fuck it

ast2vec/cloning.py
```diff
@@ -0,0 +1,178 @@
...
+
+
+class LinguistFailedError(Exception):
+    """
+    Raised when we fail to classify the source files.
+    """
+    pass
+
+
+class RepoCloner:
```
vmarkovtsev: Missing docstring

ast2vec/cloning.py
```diff
@@ -0,0 +1,178 @@
...
+    def __init__(self, linguist, languages=None, log_level=logging.INFO, num_threads=1):
+        self._log = logging.getLogger("repo_cloner")
+        self._log.setLevel(log_level)
+        self._languages = languages
+        self._linguist, self._is_enry = self._find_enry(linguist)
+        self._num_threads = num_threads
+
+    def clone_repo(self, git_url, ignore, target_dir, redownload=False):
```
vmarkovtsev: Missing docstring

ast2vec/cloning.py
```diff
@@ -0,0 +1,178 @@
...
+                            "\n\t".join(e.output.decode("utf8").split("\n")))
+            if not ignore:
+                raise e from None
+
+        self._log.info("Finished cloning %s", git_url)
+        return repo_dir
+
+    def classify_repo(self, repo_dir):
+        self._log.warning(str(os.listdir(repo_dir)))
+        if self._linguist:
```
vmarkovtsev:
```
if not self._linguist:
    raise LinguistFailedError("Linguist is not set - cannot classify the files")
```

The check must be on the caller side

ast2vec/cloning.py
```diff
@@ -0,0 +1,178 @@
...
+            self._log.error("Git failed to clone repo. git stderr:\n\t" +
+                            "\n\t".join(e.output.decode("utf8").split("\n")))
+            if not ignore:
+                raise e from None
+
+        self._log.info("Finished cloning %s", git_url)
+        return repo_dir
+
+    def classify_repo(self, repo_dir):
+        self._log.warning(str(os.listdir(repo_dir)))
```
vmarkovtsev: This is obviously the debug code you've forgotten

ast2vec/cloning.py
```diff
@@ -0,0 +1,178 @@
...
+                bjson = sb.check_output(cmdline)
+            except sb.CalledProcessError:
+                self._log.error("Couldn't classify files in %s", repo_dir)
+                raise LinguistFailedError() from None
+            classified = json.loads(bjson.decode("utf8"))
+            self._log.info("Result: %s", {k: len(v) for k, v in classified.items()})
+            return classified
+
+    def cleanup_repo(self, classified, repo_dir):
+        if classified:
```
vmarkovtsev: Please remove this. The check must be on the caller side

ast2vec/cloning.py
```diff
@@ -0,0 +1,178 @@
...
+            for dirname, dirnames, filenames in os.walk(str.encode(repo_dir)):
+                for filename in filenames:
+                    full_filename = os.path.join(dirname, filename)
+                    if full_filename not in allowed_files:
+                        os.remove(full_filename)
+
+    def clone_repos(self, inputs, output, ignore, redownload):
+        with Pool(self._num_threads) as pool:
+            pool.map(partial(self.process_repo, ignore=ignore, target_dir=output,
+                             redownload=redownload),
```
vmarkovtsev: `redownload` should be the instance's var. It fits perfectly into your **kwargs

vmarkovtsev:
\+ Please squash

vmarkovtsev:
@EgorBu @zurk I consent with you guys. Once Tim fixes the review, I need you to seek for the problems too.

ast2vec/cloning.py
fineguy: `subprocess.check_ouput` raises either `TimeoutExpired` or `CalledProcessError`. The former isn't raised if we don't set `timeout` explicitly.

vmarkovtsev: Oh, Tim, Tim... All system calls like this can throw funny and completely unexpected exceptions. For example, when we debugged the zombie threads issue with @EgorBu, the number of processes in the system reached the limit and this call raised OSError with "resource exhausted". That happened because of the way Linux spawn a new process: fork() + exec() in the child. fork() fails before exec() and voila.

vmarkovtsev: Actually, all stress situations are handled badly in Python, e.g. when the disk space ends.

ast2vec/cloning.py
```diff
@@ -0,0 +1,213 @@
...
+        try:
+            bjson = subprocess.check_output(cmdline)
+        except subprocess.CalledProcessError:
+            self._log.error("Couldn't classify files in %s", repo_dir)
+            raise LinguistFailedError() from None
+        classified = json.loads(bjson.decode("utf8"))
+        self._log.info("Result: %s", {k: len(v) for k, v in classified.items()})
+        return classified
+
+    def cleanup_repo(self, classified: dict, repo_dir: str) -> None:
```
zurk: may be this function should be protected?

ast2vec/cloning.py
```diff
@@ -0,0 +1,213 @@
...
+                           ",".join(self._languages))
+            languages = self._languages
+        else:
+            self._log.info("Removing files not classified by enry.")
+            languages = classified.keys()
+
+        allowed_files = set(str.encode(os.path.join(repo_dir, fname)) for lang in languages
+                            if lang in classified for fname in classified[lang])
+
+        for dirname, dirnames, filenames in os.walk(str.encode(repo_dir)):
```
zurk: rename dirnames to _ if you don't need this variable. [Proof](https://stackoverflow.com/questions/431866/ignore-python-multiple-return-value).

ast2vec/tests/test_cloning.py
```diff
@@ -0,0 +1,128 @@
...
+            "https://whatever.cite.com/cool/you",
+            "https://whatever.cite.com/cool/you",
+            "https://whatever.cite.com/cool/you.git",
+        ]
+        for correct_url, repo_url in zip(results, self.repo_urls):
+            self.assertEqual(correct_url, RepoCloner._prepare_repo_url(repo_url))
+
+    def _collect_repo_files(self, repo_dir):
+        files = list()
+        for dirname, dirnames, filenames in os.walk(str.encode(repo_dir)):
```
zurk: rename dirnames to _ if you don't need this variable. [Proof](https://stackoverflow.com/questions/431866/ignore-python-multiple-return-value).
🤓

ast2vec/cloning.py
```diff
@@ -0,0 +1,213 @@
...
+        """
+        bad_endings = "\n\r\\/"
+        git_url = git_url.rstrip(bad_endings)
+        if not git_url.startswith("git://") and not git_url.startswith("https://") and \
+                not git_url.startswith("http://"):
+            git_url = "https://" + git_url
+        return git_url
+
+
+def clone_repositories(args):
```
zurk: doc string

ast2vec/repo2/base.py
```diff
@@ -10,6 +10,8 @@
 import threading
 from collections import namedtuple
 
+from ast2vec.cloning import RepoCloner
+
```
zurk: change order of imports: common, external, internal.

ast2vec/repo2/base.py
```diff
@@ -47,72 +42,27 @@ def __init__(self, tempdir=None, linguist=None, log_level=logging.INFO,
...
 
-        :param url_or_path: Fiel system path to the repository or a URL to clone.
+        :param url_or_path: File system path to the repository or a URL to clone.
         :return: Some object(s) which are returned from convert_uasts().
         """
         temp = not os.path.exists(url_or_path)
         if temp:
-            target_dir = self.clone_repository(url_or_path)
+            target_dir = tempfile.mkdtemp(
+                prefix="repo2nbow-", dir=self._tempdir)
```
zurk: may be it is time to rename `repo2nbow` prefix to `repo2` prefix because it is common

ast2vec/cloning.py
fineguy: I feel that it's better to be explicit in here – in case someone wants to modify this part, or people might want to run this code in their Jupyter Notebook, which will slightly mess with the value of `_`.
Anyways, I'll probably use `dirnames` to remove empty directories as you suggested 😄

ast2vec/cloning.py
fineguy: Could you tell me why? My reasoning not to have it protected was that you might want to define your own pipeline of working with repositories (like modifying enry output in some way before cleaning up). But I seriously don't know what would be better in our situation, so I'm open to your suggestions.

vmarkovtsev: I am fine with leaving this as is

ast2vec/cloning.py
vmarkovtsev: I don't think it is really serious. I am fine with both variants. Particularly in this `os.walk` case I am always forgetting the order and get into trouble, so the present one has it's own benefits.

ast2vec/cloning.py
```diff
@@ -0,0 +1,213 @@
...
+        return classified
+
+    def cleanup_repo(self, classified: dict, repo_dir: str) -> None:
+        """
+        Delete files not classified by enry (if some languages were specified) then only them
+        are preserved.
+
+        :param classified: enry output loaded from JSON.
+        :param repo_dir: Path to repository directory.
+        :return:
```
fineguy: https://github.com/src-d/ast2vec/blob/master/ast2vec/repo2/base.py#L320

fineguy: Will use `None` for clarity.

ast2vec/cloning.py
EgorBu: +1 don't think that it's really serious)
but I think it's better to make explicit

ast2vec/tests/test_cloning.py
```diff
@@ -0,0 +1,128 @@
...
+            self.assertEqual(os.stat(enry).st_mode & 0o777, 0o777)
+            output = subprocess.check_output([enry, "-json", args.output])
+            lang_files = json.loads(output.decode("utf-8"))
+
+            files = [str.encode(os.path.join(args.output, f)) for files in
+                     lang_files.values() for f in files]
+            repo_files = self._collect_repo_files(args.output)
+            self.assertEqual(set(files), set(repo_files))
+
+            if hasattr(args, 'languages'):
```
EgorBu: quotes
@vmarkovtsev , we definitely need to share some kind of script that will check style. It's so easy to miss quotes

ast2vec/cloning.py
```diff
@@ -0,0 +1,213 @@
...
+        git_url = self._prepare_repo_url(git_url)
+
+        try:
+            repo_dir = self._prepare_repo_dir(git_url, target_dir, self._redownload)
+        except FileExistsError as e:
+            self._log.warning("%s already cloned, skipping.", git_url)
+            return
+
+        env = os.environ.copy()
+        env["GIT_TERMINAL_PROMPT"] = "0"
```
EgorBu: It's better to put explanation why some magic happens with env variable)
Like: "force git not to ask anything".

ast2vec/cloning.py
```diff
@@ -0,0 +1,213 @@
...
+        if not self._linguist:
+            raise LinguistFailedError("Linguist is not set - cannot classify the files")
+
+        self._log.info("Classifying the files...")
+        repo_dir = os.path.abspath(repo_dir)
+        cmdline = [self._linguist]
+        if self._is_enry:
+            cmdline += ["-json", repo_dir]
+        else:
+            cmdline += [repo_dir, "--json"]
```
EgorBu: I think that nothing was tested with linguist, but still, it can be used) let's keep it as it's right now

ast2vec/cloning.py
zurk: ok

ast2vec/cloning.py
zurk: ok

ast2vec/cloning.py
zurk: >Anyways, I'll probably use dirnames to remove empty directories as you suggested 😄

Good point.
Ok!

ast2vec/cloning.py
```diff
@@ -0,0 +1,213 @@
...
+                if linguist is None:
+                    linguist = "enry"
+            else:
+                return None, False
+        full_path = shutil.which(linguist)
+        if not full_path:
+            raise FileNotFoundError("%s was not found. Install it: python3 -m ast2vec enry" %
+                                    linguist)
+        with open(full_path, "rb") as fin:
+            return linguist, fin.read(15) != b"#!/usr/bin/ruby"
```
EgorBu: May you add comment about logic `fin.read(15) != b"#!/usr/bin/ruby"`?
or it's better to create parameter/attribute/etc instead of magic variable `b"#!/usr/bin/ruby"` (related to linguist)

ast2vec/repo2/base.py
```diff
@@ -20,13 +22,6 @@
                                ['filepath', 'filename', 'response'])
```
EgorBu: not related to this commits: if you can fix quotes `'' -> ""` it will be nice)

ast2vec/cloning.py
vmarkovtsev: I did test it and it used to work, though not included into the tests.

PR #32, opened 2017-07-24
Update bblfsh requirement

vmarkovtsev:


setup.py
```diff
@@ -12,7 +12,7 @@
     description="Part of source{d}'s stack for machine learning on source "
                 "code. Provides API and tools to train and use models based "
                 "on source code identifiers extracted from Babelfish's UASTs.",
-    version="0.1.0-alpha",
+    version="0.1.1-alpha",
```
zurk: I think we should also change the version in __init__.py file.
Now it is 1.0.0 for some reason.

vmarkovtsev: This is unfortunately used in model files... I hope nothing will break if I fix it.

PR #33, opened 2017-07-24
Fix output, namedtuple should not be verbose

zurk:


ast2vec/repo2/base.py
```diff
@@ -17,7 +17,7 @@
 from modelforge.model import write_model
 
 GeneratorResponse = namedtuple('GeneratorResponse',
-                               ['filepath', 'filename', 'response'], verbose=True)
```
vmarkovtsev: Oh Konst what the hell, never leave your debug code in PRs!

zurk: I missed that, sorry.

PR #35, opened 2017-07-26
Improve logging and exception catching

zurk:
This edits hard to test and it is more for our own convenience.
So I do not cover this changes with tests.

### DIFF
--- a/ast2vec/repo2/base.py
+++ b/ast2vec/repo2/base.py
@@ -1,4 +1,4 @@
-import json
+from collections import namedtuple
 import logging
 import multiprocessing
 import os
@@ -8,12 +8,11 @@ import shutil
 import subprocess
 import tempfile
 import threading
-from collections import namedtuple
 
 from bblfsh import BblfshClient
 from bblfsh.launcher import ensure_bblfsh_is_running
+from google.protobuf.message import DecodeError
 import Stemmer
-
 from modelforge.model import write_model
 
 from ast2vec.cloning import RepoCloner
@@ -75,22 +74,38 @@ class Repo2Base:
                             break
                         try:
                             dirname, filename, language = task
-                            filepath = os.path.join(dirname, filename).encode('utf8')
-                            # I need .encode('utf8') to avoid problems
-                            # with bad symbols in file names on Ubuntu
+                            filepath_naive = os.path.join(dirname, filename)
 
                             # Check if file path is symlink
-                            if os.path.islink(filepath):
-                                filepath = os.readlink(filepath)
+                            islink = os.path.islink(filepath_naive)
+                            if islink:
+                                filepath = os.readlink(filepath_naive)
+                            else:
+                                filepath = filepath_naive
+                            # sometimes you have symlink to nonexistent files. We need to check it
+                            if not os.path.exists(filepath):
+                                msg = ("File %s does not exist.\n\t%s", filepath,
+                                       "Get it from %s symlink." % filepath_naive
+                                       if islink else "And it is not a symlink...")
+                                self._log.warning(*msg)
+                                queue_out.put_nowait(None)
+                                continue
 
                             size = os.stat(filepath).st_size
                             if size > self.MAX_FILE_SIZE:
                                 self._log.warning("%s is too big - %d", filepath, size)
                                 queue_out.put_nowait(None)
                                 continue
+                            try:
+                                response = self._bblfsh[thread_index].parse(
+                                    filepath, language=language, timeout=self._timeout)
+                            except DecodeError as e:
+                                msg = "bblfsh raise an DecodeError exception. Probably your" + \
+                                      "protobuf is <= v3.3.2 and you hit https://github.com" + \
+                                      "/bblfsh/server/issues/59#issuecomment-318125752"
+                                self._log.warning(msg)
+                                raise e from None
 
-                            response = self._bblfsh[thread_index].parse(
-                                filepath, language=language, timeout=self._timeout)
                             if response is None:
                                 self._log.warning("bblfsh timed out on %s", filepath)
                             queue_out.put_nowait(GeneratorResponse(filepath=filepath,
@@ -131,7 +146,7 @@ class Repo2Base:
                     thread.join()
 
                 if empty:
-                    self._log.warning("No files were processed")
+                    self._log.warning("No files were processed for %s", url_or_path)
 
             return self.convert_uasts(file_uast_generator())
         finally:
--- a/ast2vec/repo2/base.py
+++ b/ast2vec/repo2/base.py
@@ -16,6 +16,7 @@ import Stemmer
 from modelforge.model import write_model
 
 from ast2vec.cloning import RepoCloner
+from ast2vec import resolve_symlink
 
 GeneratorResponse = namedtuple("GeneratorResponse",
                                ["filepath", "filename", "response"])
@@ -74,20 +75,13 @@ class Repo2Base:
                             break
                         try:
                             dirname, filename, language = task
-                            filepath_naive = os.path.join(dirname, filename)
+                            filepath = os.path.join(dirname, filename)
 
-                            # Check if file path is symlink
-                            islink = os.path.islink(filepath_naive)
-                            if islink:
-                                filepath = os.readlink(filepath_naive)
-                            else:
-                                filepath = filepath_naive
-                            # sometimes you have symlink to nonexistent files. We need to check it
-                            if not os.path.exists(filepath):
-                                msg = ("File %s does not exist.\n\t%s", filepath,
-                                       "Get it from %s symlink." % filepath_naive
-                                       if islink else "And it is not a symlink...")
-                                self._log.warning(*msg)
+                            try:
+                                # Resolve symlink
+                                filepath = resolve_symlink.resolve_symlink(filepath)
+                            except resolve_symlink.SymlinkToNotExistingFile as e:
+                                self._log.warning(*e.args)
                                 queue_out.put_nowait(None)
                                 continue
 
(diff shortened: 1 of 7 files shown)


ast2vec/repo2/base.py
```diff
@@ -2,18 +2,19 @@
...
 from collections import namedtuple
+from queue import Queue
 
-from bblfsh import BblfshClient
-from bblfsh.launcher import ensure_bblfsh_is_running
 import Stemmer
+from google.protobuf.message import DecodeError
 
+from bblfsh import BblfshClient
+from bblfsh.launcher import ensure_bblfsh_is_running
```
vmarkovtsev: Nope, this is external to ast2vec

ast2vec/repo2/base.py
```diff
@@ -2,18 +2,19 @@
...
 import multiprocessing
 import os
-from queue import Queue
 import re
 import shutil
 import subprocess
 import tempfile
 import threading
 from collections import namedtuple
+from queue import Queue
```
vmarkovtsev: O P Q R S T U V
move collections instead

ast2vec/repo2/base.py
```diff
@@ -126,22 +127,41 @@ def thread_loop(thread_index):
                             break
                         try:
                             folder, filename, language = task
-                            filepath = os.path.join(folder, filename).encode('utf8')
+                            filepath_naive = os.path.join(folder, filename).encode('utf8')
```
vmarkovtsev: Single quotes everywhere, bad

ast2vec/repo2/base.py
```diff
@@ -126,22 +127,41 @@ def thread_loop(thread_index):
...
                             # I need .encode('utf8') to avoid problems
                             # with bad symbols in file names on Ubuntu
```
vmarkovtsev: lol, ubuntu has nothing to do with unicode errors

ast2vec/repo2/base.py
```diff
@@ -126,22 +127,41 @@ def thread_loop(thread_index):
...
 
                             # Check if file path is symlink
-                            if os.path.islink(filepath):
-                                filepath = os.readlink(filepath)
+                            islink = os.path.islink(filepath_naive)
+                            if islink:
+                                filepath = os.readlink(filepath_naive)
+                            else:
+                                filepath = filepath_naive
+                            if not os.path.exists(filepath):
```
vmarkovtsev: You need to put a comment here - not obvious what is happening

ast2vec/repo2/base.py
```diff
@@ -126,22 +127,41 @@ def thread_loop(thread_index):
...
+                                self._log.warning('File %s does not exist.\n\t%s',
```
vmarkovtsev: Is it really 99 chars limit? Damn. Looks bad

Quotes

ast2vec/repo2/base.py
```diff
@@ -126,22 +127,41 @@ def thread_loop(thread_index):
...
+                                                  'Get it from %s symlink.' %
+                                                  filepath_naive.decode('utf8')
+                                                  if islink else 'And it is not a symlink...')
+                                queue_out.put_nowait(None)
+                                continue
 
                             size = os.stat(filepath).st_size
                             if size > self.MAX_FILE_SIZE:
-                                self._log.warning("%s is too big - %d", filepath, size)
+                                self._log.warning("%s is too big - %d",
```
vmarkovtsev: Not needed

ast2vec/repo2/source.py
```diff
@@ -23,7 +23,14 @@ def convert_uasts(self, file_uast_generator):
         filenames = []
 
         for file_uast in file_uast_generator:
-            src_codes.append(self._get_sources(file_uast.filepath))
+            try:
+                sources = self._get_sources(file_uast.filepath)
+            except UnicodeDecodeError as e:
+                # We can code a simple tool to get python file encoding later.
```
vmarkovtsev: We will not. We will base64 the read bytes

ast2vec/repo2/base.py
```diff
@@ -126,22 +127,41 @@ def thread_loop(thread_index):
...
+                                                  filepath.decode('utf8'), size)
                                 queue_out.put_nowait(None)
                                 continue
+                            try:
+                                response = self._bblfsh[thread_index].parse(
+                                    filepath, language=language, timeout=self._timeout)
+                            except DecodeError as e:
+                                self._log.warning("bblfsh raise an DecodeError exception. "
+                                                  "Usually it is because protobuf package v3.3.0. "
```
vmarkovtsev: Probably your protobuf is >= v3.3.0 and you hit https://github.com/bblfsh/server/issues/59#issuecomment-318125752

ast2vec/repo2/base.py
vmarkovtsev: Why do we convert this to bytes?

ast2vec/repo2/base.py
zurk: Ok, there is nothing about the order of `import ...` and `from ... import ...` in PEP8.
My style feature. I change it.

zurk: but also I can not find anything about the alphabetical order of imports. 
And find an awesome answer on [StackOverflow](https://stackoverflow.com/a/20763446).
I think we need discuss it and add to our style gide.

ast2vec/repo2/base.py
zurk: yes, I do not like it too.
I think to put it out from such a deep nesting, but not sure it is a good idea.

ast2vec/repo2/base.py
zurk: If I set logging level to `WARNING` it was unclear about such warnings.
And `INFO` level is too verbose now.

ast2vec/repo2/base.py
zurk: Yeah, It is my comment :)

ast2vec/repo2/base.py
vmarkovtsev: It was discussed and unfortunately missed because I did not check Alfredo: https://github.com/src-d/guide/pull/52#discussion_r129808242

ast2vec/repo2/base.py
vmarkovtsev: I mean, try to not change the lines if you do not have to. This is a bad contribution style, so called "white knight", it slows me down on reviews and worsens blaming.

ast2vec/repo2/base.py
vmarkovtsev: I mean change the comment exactly how it is written by me

ast2vec/repo2/base.py
vmarkovtsev: Try to refactor this into a method?

ast2vec/repo2/base.py
zurk: For example [this repo](https://github.com/ROBelgium/MSNoise) has bad filename:
https://github.com/ROBelgium/MSNoise/blob/master/msnoise/unused/%C3%A903_prewhitened.py
And If I remove `.encode('utf8')` I have
```
ERROR:repo2source:Error while processing ('/tmp/repo2-r27femkp/ROBelgium@MSNoise', 'msnoise/unused/é03_prewhitened.py', 'Python')
Traceback (most recent call last):
  File "/usr/local/lib/python3.5/dist-packages/ast2vec/repo2/base.py", line 86, in thread_loop
    islink = os.path.islink(filepath_naive)
  File "/usr/lib/python3.5/posixpath.py", line 161, in islink
    st = os.lstat(path)
UnicodeEncodeError: 'ascii' codec can't encode character '\xe9' in position 53: ordinal not in range(128)
```
And I have this problem only under Ubuntu, not MacOS.
Os function can not deal with Unicode string (may be because of locality or whatever) and the simplest way to avoid this problem use byte sequence.

I check it one time again. So, I should keep it.
If you want I can make str filepath and byte filepath just to avoid call .decode('utf8') each time in logging.

ast2vec/repo2/base.py
zurk: Mmm... but I have to because it was hard to debug with this log message. 
Should I change them in this case?

ast2vec/repo2/base.py
zurk: Ok, I will push it now with new edits and after your approvement refactor it.

ast2vec/repo2/base.py
vmarkovtsev: Strike me lightning, you just added the newline character here

ast2vec/repo2/base.py
vmarkovtsev: It has nothing to do with Ubuntu but rather Linux fs in general or the specific environment. In Linux world, distribution rarely means anything: the packages are built from the same source. The rest is the environment settings

ast2vec/repo2/base.py
vmarkovtsev: So remove the damn \n and leave it as-is :)

vmarkovtsev:
I see a dirty series here, rebase and squash

ast2vec/repo2/base.py
```diff
@@ -74,20 +75,13 @@ def thread_loop(thread_index):
...
-                                msg = ("File %s does not exist.\n\t%s", filepath,
-                                       "Get it from %s symlink." % filepath_naive
-                                       if islink else "And it is not a symlink...")
-                                self._log.warning(*msg)
+                            filepath = os.path.join(dirname, filename)
+
+                            try:
+                                # Resolve symlink
+                                filepath = utils.solve_symlink(filepath)
+                            except utils.SymlinkToNotExistingFile as e:
```
vmarkovtsev: Always end your exceptions with "Error". here: "DanglingSymlinkError" (because such broken symlinks are called dangling)

ast2vec/utils.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+class SymlinkToNotExistingFile(Exception):
+    """
+    Exception for solve_symlink function when symlink converts to nonexistent file
+    """
+    pass
+
+
+def solve_symlink(path: str) -> str:
```
vmarkovtsev: resolve

ast2vec/utils.py
```diff
@@ -0,0 +1,26 @@
+import os
```
vmarkovtsev: Please rename this file to resolve_symlink.py

ast2vec/utils.py
```diff
@@ -0,0 +1,26 @@
...
+class SymlinkToNotExistingFile(Exception):
+    """
+    Exception for solve_symlink function when symlink converts to nonexistent file
+    """
+    pass
+
+
+def solve_symlink(path: str) -> str:
+    """
+    Resolve symlink if path is symlink
```
vmarkovtsev: ...is a symbolic link.

ast2vec/utils.py
```diff
@@ -0,0 +1,26 @@
...
+    Check file existent. If it does not exist raise SymlinkToNotExistingFile Exception
```
vmarkovtsev: existence + do not forget to update the exception name here

PR #45, opened 2017-08-07
Add token parser

fineguy:


ast2vec/token_parser.py
```diff
@@ -0,0 +1,63 @@
+import re
+
+import Stemmer
+
+
+class TokenParser:
+    """
+    Token parser.
```
vmarkovtsev: Come on :) Describe what it does in more detail

PR #71, opened 2017-08-17
Fix memory leak

fineguy:


ast2vec/model2/proxbase.py
```diff
@@ -135,17 +111,37 @@ def fill_mat(item_to_item, offset):
...
+        :param mat: Adjacency matrix ('scipy.sparse.coo_matrix') with rows corresponding to
+                    node roles followed by node tokens.
+        :return: tuple('tokens', 'matrix'). 'tokens' are generalized tokens (usually roles+tokens).
+                 'matrix' rows correspond to 'tokens'.
+        """
+        raise NotImplementedError
+
+    def _clear(self):
+        """
+        Release memory.
```
vmarkovtsev: You call this from __init__ and apparently do not release any memory in that case. Rephrase

PR #88, opened 2017-09-12
Reanimate the tests

vmarkovtsev:
Reanimate the tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,6 +36,8 @@ matrix:
       after_success:
         - codecov
   fast_finish: true
+before_script:
+- docker run --rm --privileged -d -p 9432:9432 --name bblfsh bblfsh/server:v0.7.0
 script:
 - (eval "$SCRIPT")
 notifications:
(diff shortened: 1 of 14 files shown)


vmarkovtsev:
That was tricky.

I changed the `multiprocessing` launch method to `forkserver`. This means there is a "zygote" Python interpreter and each time we need a worker, we fork from it and set it up. It is slower than normal launch method - "fork" - but faster than "spawn" which launches a fresh python interpreter. This is the only way to ensure that we don't have grpc imported beforehand.

@zurk Future improvement: please move the launch method setting from `base.py` to `tests/__init__.py` and ensure that `process_entry()` calls `sys.exit()` instead of `os_exit()` in that case.

I had to fix a few early grpc imports. Since some modules require grpc during import, e.g. `bblfsh_roles`, I went hardcore and implemented the "masquerade".

@zurk I had to skip your tests in `test_repo2source.py`. They are broken because the new launch method is incompatible with your monkey patching. Please invent smth.

I hope we will never struggle with this gentlemen.

PR #93, opened 2017-09-18
Add time logging

zurk:
Need to add some tests later.

### DIFF
--- a/ast2vec/repo2/base.py
+++ b/ast2vec/repo2/base.py
@@ -1,6 +1,7 @@
 from collections import namedtuple
 from itertools import accumulate, repeat
 import logging
+from math import isnan
 import multiprocessing
 import os
 from queue import Queue
@@ -12,6 +13,7 @@ import tempfile
 import threading
 from typing import Union
 from time import time
+import tempfile
 
 from google.protobuf.message import DecodeError
 from modelforge.logs import setup_logging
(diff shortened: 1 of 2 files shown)


vmarkovtsev:
Please add some tests - delta is 1%

ast2vec/repo2/base.py
```diff
@@ -166,15 +182,39 @@ def convert_uasts(self, file_uast_generator):
 
     def _bblfsh_parse(self, thread_index, filepath, language):
         from grpc import RpcError
+        status = "ERR"
+        start_time = time()
+        lines = lines_count(filepath, self._log)
```
vmarkovtsev: Please calc the lines only if we need

ast2vec/repo2/base.py
```diff
@@ -166,15 +182,39 @@ def convert_uasts(self, file_uast_generator):
...
                 filepath, language=language, timeout=self._timeout)
+            status = "OK"
+            return res
         except DecodeError as e:
             msg = "bblfsh: DecodeError on %s: %s\nYour protobuf may be <= v3.3.2 " \
                   "and you hit https://github.com/bblfsh/server/issues/59#issuecomment-318125752"
             self._log.error(msg, e)
         except RpcError as e:
             self._log.error("bblfsh: RpcError on %s: %s", filepath, e)
+            self._log.warning(msg)
```
vmarkovtsev: error and warning in the same place is too much. Besides, there is no `msg`. Mismerge?

ast2vec/repo2/base.py
```diff
@@ -166,15 +182,39 @@ def convert_uasts(self, file_uast_generator):
...
+            raise e from None
+        finally:
+            self._write_debug_logs(thread_index, filepath, status, lines, start_time)
+
+    def _write_debug_logs(self, thread_index, filepath, status, lines, start_time):
+        time_spent = time() - start_time
+        if not isnan(lines):
+            self._lines += lines
+            self._time += time_spent
+        if self._log.isEnabledFor(logging.DEBUG):
```
vmarkovtsev: Please check an environment variable instead. I think that if you do not define `AST2VEC_PARSE_LOG` then we should not write anything. This implies there is no `DEFAULT_PARSE_LOG_FILENAME `

ast2vec/repo2/base.py
```diff
@@ -166,15 +182,39 @@ def convert_uasts(self, file_uast_generator):
...
+            with lock:
+                if not os.path.exists(self.PARSE_LOG_FILENAME):
```
vmarkovtsev: Please write this initial header somewhere in the very beginning, before any multiprocessing. Otherwise, several processes may be in a race.

vmarkovtsev:
All good! Just a minor thing.

zurk:
fixed

PR #101, opened 2017-09-20
Add python module installation to `ast2vec bigartm` command

zurk:


ast2vec/bigartm.py
```diff
@@ -51,6 +51,7 @@ def install_bigartm(args=None, target="./bigartm", tempdir=None, warn_exists=Tru
         execute("cmake -DCMAKE_BUILD_TYPE=Release -DBUILD_TESTING=OFF -DPYTHON=python3 ..",
                 cwd, log)
         execute("make -j%d" % multiprocessing.cpu_count(), cwd, log)
+        execute("make install", cwd, log)
```
vmarkovtsev: Nopes. It must be `pip3 install path/to/wheel` otherwise it is installed without the way to remove it.

PR #103, opened 2017-09-29
Update ast2vec for bblfsh version 1.0.1

zurk:


vmarkovtsev:
Thanks for the contribution. I used it in the massive refactoring branch "engine" which removes repo2* and model2* in favor of source{d} engine.

PR #104, opened 2017-10-02
Fix organize_files flag

fineguy:
I.e. you had `git://github.com/author/repo.git` it would expand into `g/github.com&author&repo.asdf` instead of `a/au/github.com&author&repo.asdf`.


vmarkovtsev:
No longer actual, sadly, since we switched to source{d} engine.

PR #114, opened 2017-11-15
Fix all after ast2vec renaming to sourced.ml

zurk:


Dockerfile
```diff
@@ -26,8 +25,15 @@ echo\n\' > /browser && \
 
 RUN pip3 install --no-cache-dir ./package && rm -rf package
 
+RUN wget "https://github.com/src-d/enry/releases/download/v1.6.1/enry_v1.6.1_linux_amd64.tar.gz" && \
```
vmarkovtsev: we have `sourcedml enry` exactly for checking out the latest release

zurk: we dont have `sourcedml enry` you delete it :) 
https://github.com/src-d/ml/commit/22f0350b279a645f7609548679f6b22cb5d57592#diff-7a0d9bbf1d938a3c2b74d90b7d76d6f6
mistake?

vmarkovtsev: Now I see! We don't need enry anymore. Just remove it :)

zurk: oh, yes of course!

PR #125, opened 2017-11-27
[RFC] Implement repos2coocc on source{d} engine

zurk:
Here the first implementation of repos2coocc. 
It is raw and It needs more comments, tests.
But still, one can take a look and give useful feedback.

RFC = request for comments.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,4 +5,4 @@ modelforge>=0.5.0-alpha
 parquet==1.2
 numpy>=1.12,<2.0
 pyspark==2.2.0
-sourced-engine==0.1.10
+sourced-engine==0.2.0,<0.3.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,3 +2,7 @@ PyStemmer>=1.3,<2.0
 tensorflow>=1.0,<2.0
 bblfsh>=2.2.1,<3.0
 modelforge>=0.5.0-alpha
+parquet==1.2
+numpy>=1.12,<2.0
+pyspark==2.2.0
+sourced-engine==0.1.10
--- a/setup.py
+++ b/setup.py
@@ -29,9 +29,8 @@ setup(
                       "bblfsh>-2.2.1,<3.0",
                       "modelforge>=0.5.0-alpha",
                       "parquet==1.2",
-                      "numpy>=1.12,<2.0",
                       "pyspark==2.2.0",
-                      "sourced-engine == 0.1.10"] + typing,
+                      "sourced-engine==0.2.0"] + typing,
     extras_require={
         "tf": ["tensorflow>=1.0,<2.0"],
         "tf_gpu": ["tensorflow-gpu>=1.0,<2.0"],
--- a/setup.py
+++ b/setup.py
@@ -27,7 +27,11 @@ setup(
               "github", "swivel", "nbow", "bblfsh", "babelfish"],
     install_requires=["PyStemmer>=1.3,<2.0",
                       "bblfsh>-2.2.1,<3.0",
-                      "modelforge>=0.5.0-alpha"] + typing,
+                      "modelforge>=0.5.0-alpha",
+                      "parquet==1.2",
+                      "numpy>=1.12,<2.0",
+                      "pyspark==2.2.0",
+                      "sourced-engine == 0.1.10"] + typing,
     extras_require={
         "tf": ["tensorflow>=1.0,<2.0"],
         "tf_gpu": ["tensorflow-gpu>=1.0,<2.0"],
--- a/sourced/ml/__main__.py
+++ b/sourced/ml/__main__.py
@@ -4,6 +4,7 @@ import os
 import sys
 
 from sourced.ml.bigartm import install_bigartm
+from sourced.ml.engine import SparkDefault, EngineDefault
 from sourced.ml.id2vec import projector_entry
 from sourced.ml.id_embedding import preprocess as preprocess_id2vec, run_swivel, \
     postprocess as postprocess_id2vec, swivel
@@ -40,9 +41,11 @@ def one_arg_parser(*args, **kwargs) -> argparse.ArgumentParser:
 
 def add_spark_args(my_parser):
     my_parser.add_argument(
-        "-s", "--spark", default="local[*]", help="Spark's master address.")
+        "-s", "--spark", default=SparkDefault.MASTER_ADDRESS,
+        help="Spark's master address.")
     my_parser.add_argument(
-        "--config", nargs="+", default=[], help="Spark configuration (key=value).")
+        "--config", nargs="+", default=SparkDefault.CONFIG,
+        help="Spark configuration (key=value).")
     my_parser.add_argument(
         "-m", "--memory",
         help="Handy memory config for spark. -m 4,10,2 is equivalent to "
@@ -51,9 +54,11 @@ def add_spark_args(my_parser):
              "--config spark.driver.maxResultSize=2G."
              "Numbers are floats separated by commas.")
     my_parser.add_argument(
-        "--package", nargs="+", default=[], help="Additional Spark package.")
+        "--package", nargs="+", default=SparkDefault.PACKAGE,
+        help="Additional Spark package.")
     my_parser.add_argument(
-        "--spark-local-dir", default="/tmp/spark", help="Spark local directory.")
+        "--spark-local-dir", default=SparkDefault.LOCAL_DIR,
+        help="Spark local directory.")
     persistences = ("DISK_ONLY", "DISK_ONLY_2", "MEMORY_ONLY", "MEMORY_ONLY_2",
                     "MEMORY_AND_DISK", "MEMORY_AND_DISK_2", "OFF_HEAP")
     my_parser.add_argument(
@@ -64,9 +69,11 @@ def add_spark_args(my_parser):
 def add_engine_args(my_parser):
     add_spark_args(my_parser)
     my_parser.add_argument(
-        "--bblfsh", default="localhost", help="Babelfish server's address.")
+        "--bblfsh", default=EngineDefault.BBLFSH,
+        help="Babelfish server's address.")
     my_parser.add_argument(
-        "--engine", default="0.2.0", help="source{d} engine version.")
+        "--engine", default=EngineDefault.VERSION,
+        help="source{d} engine version.")
     my_parser.add_argument("--explain", action="store_true",
                            help="Print the PySpark execution plans.")
 
--- a/sourced/ml/__main__.py
+++ b/sourced/ml/__main__.py
@@ -9,6 +9,7 @@ from sourced.ml.id_embedding import preprocess as preprocess_id2vec, run_swivel,
     postprocess as postprocess_id2vec, swivel
 from sourced.ml.topics import bigartm2asdf_entry
 from sourced.ml.vw_dataset import bow2vw_entry
+from sourced.ml.repo2.coocc import repos2coocc_entry
 
 from modelforge.logs import setup_logging
 from sourced.ml.dump import dump_model
@@ -35,6 +36,25 @@ def one_arg_parser(*args, **kwargs) -> argparse.ArgumentParser:
     arg_parser = argparse.ArgumentParser(add_help=False)
     arg_parser.add_argument(*args, **kwargs)
     return arg_parser
+
+
+def add_spark_args(my_parser):
+    my_parser.add_argument(
+        "-s", "--spark", default="local[*]", help="Spark's master address.")
+    my_parser.add_argument(
+        "--config", nargs="+", default=[], help="Spark configuration (key=value).")
+    my_parser.add_argument(
+        "--package", nargs="+", default=[], help="Additional Spark package.")
+    my_parser.add_argument(
+        "--spark-local-dir", default="/tmp/spark", help="Spark local directory.")
+
+
+def add_engine_args(my_parser):
+    add_spark_args(my_parser)
+    my_parser.add_argument(
+        "--bblfsh", default="localhost", help="Babelfish server's address.")
+    my_parser.add_argument(
+        "--engine", default="0.1.10", help="source{d} engine version.")
 
 
 def get_parser() -> argparse.ArgumentParser:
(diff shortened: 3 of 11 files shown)


requirements.txt
```diff
@@ -2,3 +2,7 @@ PyStemmer>=1.3,<2.0
 tensorflow>=1.0,<2.0
 bblfsh>=2.2.1,<3.0
 modelforge>=0.5.0-alpha
+parquet==1.2
+numpy>=1.12,<2.0
```
vmarkovtsev: numpy is included into modelforge

requirements.txt
```diff
@@ -2,3 +2,7 @@ PyStemmer>=1.3,<2.0
...
+pyspark==2.2.0
+sourced-engine==0.1.10
```
vmarkovtsev: We should use >=0.2.0, <0.3.0

setup.py
```diff
@@ -27,7 +27,11 @@
               "github", "swivel", "nbow", "bblfsh", "babelfish"],
     install_requires=["PyStemmer>=1.3,<2.0",
                       "bblfsh>-2.2.1,<3.0",
-                      "modelforge>=0.5.0-alpha"] + typing,
+                      "modelforge>=0.5.0-alpha",
+                      "parquet==1.2",
+                      "numpy>=1.12,<2.0",
+                      "pyspark==2.2.0",
+                      "sourced-engine == 0.1.10"] + typing,
```
vmarkovtsev: Same here

sourced/ml/__main__.py
```diff
@@ -37,6 +38,25 @@ def one_arg_parser(*args, **kwargs) -> argparse.ArgumentParser:
...
 
+def add_spark_args(my_parser):
+    my_parser.add_argument(
+        "-s", "--spark", default="local[*]", help="Spark's master address.")
+    my_parser.add_argument(
+        "--config", nargs="+", default=[], help="Spark configuration (key=value).")
+    my_parser.add_argument(
+        "--package", nargs="+", default=[], help="Additional Spark package.")
+    my_parser.add_argument(
+        "--spark-local-dir", default="/tmp/spark", help="Spark local directory.")
```
vmarkovtsev: Please include a handy memory config:

```
--config spark.executor.memory=4G --config spark.driver.memory=10G --config spark.driver.maxResultSize=4G
```

should be replaced with

```
-m / --memory 4,10,4
```

Numbers are floats separated by commas.

sourced/ml/__main__.py
```diff
@@ -37,6 +38,25 @@ def one_arg_parser(*args, **kwargs) -> argparse.ArgumentParser:
...
+
+
+def add_engine_args(my_parser):
+    add_spark_args(my_parser)
+    my_parser.add_argument(
+        "--bblfsh", default="localhost", help="Babelfish server's address.")
+    my_parser.add_argument(
+        "--engine", default="0.1.10", help="source{d} engine version.")
```
vmarkovtsev: The default is `0.2.0` now

sourced/ml/__main__.py
```diff
@@ -70,6 +90,30 @@ def get_parser() -> argparse.ArgumentParser:
     # Create and construct subparsers
 
     subparsers = parser.add_subparsers(help="Commands", dest="command")
+
+    repo2coocc_parser = subparsers.add_parser(
+        "repo2coocc", help="Produce the co-occurrence matrix from a Git repository.",
+        formatter_class=ArgumentDefaultsHelpFormatterNoNone,
+        parents=[])
+    add_engine_args(repo2coocc_parser)
```
vmarkovtsev: Let's be consistent. If we switch from `parents` to functions then we should refactor all the previous `parents`

sourced/ml/__main__.py
```diff
@@ -70,6 +90,30 @@ def get_parser() -> argparse.ArgumentParser:
...
+
+    repo2coocc_parser.add_argument(
+        "--min-docfreq", default=1, type=int,
+        help="The minimum document frequency of each element.")
+
+    repo2coocc_parser.add_argument(
+        "-l", "--languages", required=True, nargs="+", choices=("Java", "Python"),
+        help="The programming languages to analyse.")
+    repo2coocc_parser.add_argument(
+        "--persist", default=None, help="Persistence type (StorageClass.*).")
```
vmarkovtsev: Pls list the choices from https://github.com/apache/spark/blob/master/python/pyspark/storagelevel.py#L52 and rename Class to Level

sourced/ml/engine.py
```diff
@@ -30,3 +31,15 @@ def create_engine(session_name, repositories, kwargs):
     log.info("Initializing on %s", repositories)
     engine = Engine(session, repositories)
     return engine
+
+
+def get_tokens(uasts):
```
vmarkovtsev: No, this should go here. This file contains only the functions to launch the engine and its dependencies. Tokens extraction is a super narrow problem.

Besides, this duplicates with `uast_ids_to_bag.py` and not needed.

sourced/ml/repo2/base.py
```diff
@@ -4,6 +4,7 @@
 
 from pyspark import StorageLevel
 from pyspark.sql.types import Row
+from pyspark.sql.functions import col, udf
```
vmarkovtsev: `col` is a very widespread name. I suggest to `from pyspark.sql import functions`

sourced/ml/repo2/coocc.py
```diff
@@ -0,0 +1,131 @@
+import logging
+import os
+from operator import add
```
vmarkovtsev: Let's `import operator`

sourced/ml/repo2/coocc.py
```diff
@@ -0,0 +1,131 @@
...
+from uuid import uuid4
+from pyspark.sql import column
+from pyspark.sql.functions import size, col
```
vmarkovtsev: The same here

sourced/ml/repo2/coocc.py
```diff
@@ -0,0 +1,131 @@
...
+
+from modelforge.meta import generate_meta
+from modelforge.model import merge_strings, write_model
+from sourced.ml.engine import create_engine, get_tokens
+from sourced.ml.repo2.base import UastExtractor, Transformer, Cacher, UastDeserializer
+from sourced.ml.repo2.token_map import TokenMapTransformer
+from sourced.ml.token_parser import TokenParser
+
+import itertools
```
vmarkovtsev: Out of order

sourced/ml/repo2/coocc.py
```diff
@@ -0,0 +1,131 @@
...
+
+from sourced.ml.bblfsh_meta import Roles, Node
```
vmarkovtsev: Out of order

sourced/ml/repo2/coocc.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+class SaveASDFCooccModel(Transformer):
```
vmarkovtsev: Remove "asdf" - it is an implementation detail. Change the rest: `CooccModelSaver` - class names must be nouns.

sourced/ml/repo2/coocc.py
```diff
@@ -0,0 +1,131 @@
...
+        self.tokens_list = tokens_list
+        self.output = output
+
+    def __call__(self, sparce_matrix):
+        matrix_count = sparce_matrix.count()
+        rows = sparce_matrix.take(matrix_count)
+
+        mat_row, mat_col, mat_weights = zip(*rows)
+
+        write_model(generate_meta("co-occurrences", (1, 0, 0)),
```
vmarkovtsev: `write_model` should not be used. `Model().construct(...).save(self.output)`

sourced/ml/repo2/coocc.py
```diff
@@ -0,0 +1,131 @@
...
+                    {"tokens": merge_strings(self.tokens_list),
+                     "matrix": {
+                         "shape": (len(self.tokens_list), len(self.tokens_list)),
+                         "format": "coo",
+                         "data": (mat_weights, (mat_row, mat_col))}
+                     },
+                    self.output)
+
+
+class UASTCooccTransformer(Transformer):
```
vmarkovtsev: Pls remove "Transformer". `CooccCalculator` or smth like that.

sourced/ml/repo2/coocc.py
```diff
@@ -0,0 +1,131 @@
...
+    def __init__(self, token2index, token_parser, prune_size=1, **kwargs):
+        super().__init__(**kwargs)
+        self.token2index = token2index
+        self.token_parser = token_parser
+
+        # TODO(zurk): implement pruning
+        self.prune_size = prune_size
+
+    def _flatten_children(self, root):
+        ids = []
```
vmarkovtsev: This is replaced with `uast_ids_to_bag.py`

sourced/ml/repo2/coocc.py
```diff
@@ -0,0 +1,131 @@
...
+        stack = [uast]
+        new_stack = []
+
+        while stack:
+            for node in stack:
+                children = self._flatten_children(node)
+                tokens = []
+                for ch in children:
+                    tokens.extend(self.token_parser(ch.token))
+                if (node.token.strip() is not None and node.token.strip() != "" and
```
vmarkovtsev: I am crying tears. I guess this is Egor's legacy. Please make it sexy.

sourced/ml/repo2/coocc.py
```diff
@@ -0,0 +1,131 @@
...
+                yield (self.token2index.value[token1], self.token2index.value[token2]), 1
+            except KeyError:
+                # Do not have token1 or token2 in the token2index map
+                pass
+
+def repos2coocc_entry(args):
+    log = logging.getLogger("repos2cooc")
+    if not args.config:
+        args.config = []
+    engine = create_engine("repos2cooc-%s" % uuid4(), args.repositories, args)
```
vmarkovtsev: I have a feeling we need to refactor create_engine() and create_spark() to accept **named** `**kwargs` instead of `args`. **named** Otherwise I used to code `create_engine()` a few times and had to hack with `argparse.Namespace` to make it work, and then spent a few times figuring out which parameters were obligatory.

```
create_engine("", **args.__dict__)
```

sourced/ml/repo2/coocc.py
```diff
@@ -0,0 +1,131 @@
...
+
+    pipeline = UastExtractor(engine, languages=args.languages)
+    if args.persist is not None:
```
vmarkovtsev: Let's code

```
class Cacher:
    @staticmethod
    def maybe(pipeline, persistence):
        if persistence is not None:
            return pipeline.link(...)
        else
            return pipeline

pipeline = Cacher.maybe(pipeline, args.persistence)
```

sourced/ml/repo2/coocc.py
```diff
@@ -0,0 +1,131 @@
...
+        uasts = pipeline.link(Cacher(args.persist))
+    else:
+        uasts = pipeline
+
+    token_parser = TokenParser()
+    token_mapping_transformer = TokenMapTransformer(token_parser)
+    tokens, tokens2index = uasts.link(token_mapping_transformer).execute()
```
vmarkovtsev:
```
tokens, tokens2index = uasts.link(TokenMapTransformer(token_parser)).execute()
```

sourced/ml/repo2/token_map.py
```diff
@@ -0,0 +1,29 @@
+from operator import add
+
+from sourced.ml.engine import get_tokens
+from sourced.ml.repo2.base import Transformer
+from sourced.ml.token_parser import TokenParser
+
+
+class TokenMapTransformer(Transformer):
```
vmarkovtsev: TokenMapper?

sourced/ml/repo2/token_map.py
```diff
@@ -0,0 +1,29 @@
...
+        self.token_parser = token_parser
+        self.prune_size = prune_size
+
+    def __call__(self, uasts):
+        """
+        Make tokens list and token2index mapping from provided uasts.
+        token2index is broadcasted dictionary to use it in workers. Maps token to its index.
+        """
+        tokens = get_tokens(uasts).rdd \
+            .flatMap(lambda r: [(t, 1) for token in r.tokens for t in self.token_parser(token)]) \
```
vmarkovtsev: Can be optimized with `countByKey()`

sourced/ml/engine.py
zurk: I will delete it, but `uast_ids_to_bag.py` is outdated. Will reimplement it using new engine later.

sourced/ml/repo2/coocc.py
zurk: No, it is not the same. It was in ast2vec as separate thing and implement https://github.com/src-d/ml/blob/master/ast2vec/repo2/coocc.md. Code was here 
https://github.com/src-d/ml/blob/master/ast2vec/repo2/cooccbase.py#L50

sourced/ml/repo2/coocc.py
zurk: Have an idea how to implement it.

sourced/ml/repo2/token_map.py
zurk: depends from point of view. 
it returns python dict, and then we do pruning. So, it is much better to use this approach for master node memory since we have a huge number of rare tokens. 

Prefer to keep as is.

sourced/ml/engine.py
vmarkovtsev: It is not outdated. It contains the single payload class `UastIds2Bag` and we use it in many places. Please do not touch it.

sourced/ml/repo2/coocc.py
vmarkovtsev: Please study uast_ids_to_bag.py and remove this completely - it provides the needed generalization. If no idea how, I can show next week.

vmarkovtsev:
@zurk Regarding `uast_ids_to_bag.py`: perhaps we should rename that file - it can do any xpath query, not only ids.

sourced/ml/repo2/token_map.py
vmarkovtsev: You did not get me. `reduceByKey(add)` is exactly the same as `countByKey()` but slower. Everything else is left untouched.

zurk:
Ok, I will keep it in mind until refactor stage.

sourced/ml/repo2/token_map.py
zurk: I get what you mean, but no, it is not the same.

if I run 
```python
self.get_tokens(uasts).rdd \
            .flatMap(lambda r: [(t, 1) for token in r.tokens for t in self.token_parser(token)]) \
            .reduceByKey(add)
```
I will get the `rdd`, nothing is executed.

If I run 
```python
self.get_tokens(uasts).rdd \
            .flatMap(lambda r: [(t, 1) for token in r.tokens for t in self.token_parser(token)]) \
            .countByKey()
```
I will get python `dict`. I Do not know why it has such behavior, but I don't what to have dict here.

Just now I check it one time more.

vmarkovtsev: Yes, you are right. From official docs:

```
Count the number of elements for each key, and return the result to the master as a dictionary.
```

Indeed, if you need an rdd, this does not work. Sorry for misleading.

zurk:
I add several todos to resolve it in next PRs.
Most parts of comments are resolved.
@vmarkovtsev please take a look.

vmarkovtsev:
@zurk Can we close this?

PR #142, opened 2017-12-18
Add extractors with feature engineering

warenlg:
Corresponding issue : 
* https://github.com/src-d/backlog/issues/1100
* https://github.com/src-d/backlog/issues/1142

The first commit includes 2 different new extractors where the features are : 
* (interanl_type, number of children)
* (roles, number of children)
Both with quantization of the number of children

### DIFF
--- a/sourced/ml/algorithms/uast2nodes.py
+++ b/sourced/ml/algorithms/uast2nodes.py
@@ -1,0 +1,131 @@
+from collections import defaultdict
+import numpy
+
+from sourced.ml.algorithms.uast_ids_to_bag import FakeVocabulary
+
+
+class UastNode2Bag():
+    """
+    Converts a UAST to a bag of features that are node specific.
+    The features can be either (internal types, number of children)
+    or (roles, number of children)
+    """
+    def __init__(self, type2ind=None, roles2ind=None, children2ind=None):
+        """
+        :param type2ind: The mapping from internal types to bag of keys. \
+            If None, no mapping is performed.
+        :param roles2ind: The mapping from roles to bag of keys.
+        :param children2ind: The mapping from the number of children to bag of keys. \
+            :class:'Quantization' is used if it is not specified.
+        """
+        self.type2ind = FakeVocabulary() if type2ind is None else type2ind
+        self.roles2ind = FakeVocabulary() if roles2ind is None else roles2ind
+        self.children2ind = Quantization() if children2ind is None else children2ind
+
+    def _uast2nodes(self, root):
+        """
+        :param uast: The UAST root node.
+        :return: The list of the nodes composing the UAST.
+        """
+        nodes = []
+        queue = [root]
+        while queue:
+            child = queue.pop(0)
+            queue.extend(child.children)
+            nodes.append(child)
+        return nodes
+
+    def _nodes2children(self, nodes):
+        """
+        :param nodes: List of all nodes from a UAST.
+        :return: The list of the number of children of all these nodes.
+        """
+        nb_children = []
+        for n in nodes:
+            nb_children.append(len(n.children))
+        return nb_children
+
+    def uast_to_bag(self, uast, feature):
+        """
+        Converts a UAST to a bag-of-features. The weights are feature frequencies.
+        The number of children are preprocessed by :class:`Quantization`
+
+        :param uast: The UAST root node.
+        :param feature: The first feature we want to extract from the nodes of the UAST.
+        :return:
+        """
+        bag = defaultdict(int)
+        nodes = self._uast2nodes(uast)
+        nb_children = self._nodes2children(nodes)
+        self.children2ind.build_partition(nb_children)
+        for n in nodes:
+            bag[self.node2ind(n, feature)] += 1
+        return bag
+
+    def node2ind(self, node, feature):
+        """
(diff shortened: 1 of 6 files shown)


sourced/ml/algorithms/uast2nodes.py
```diff
@@ -0,0 +1,131 @@
...
+        :param node: a node of UAST
+        :param feature: The first feature we want to extract from the nodes of the UAST. \
+        This first feature should be either "internal_type" or "roles", \
+        the second being the number of children anyway.
+        :return:
+        """
+        if feature is "internal_type":
+            return " ".join([self.type2ind[node.internal_type],
+                             str(self.children2ind.process_value(len(node.children)))])
+        elif feature is "roles":
```
zurk: Since it is different features, I think it is better to separate it into different classes.

sourced/ml/algorithms/uast2nodes.py
```diff
@@ -0,0 +1,131 @@
+from collections import defaultdict
+import numpy
+
+from sourced.ml.algorithms.uast_ids_to_bag import FakeVocabulary
+
+
+class UastNode2Bag():
```
vmarkovtsev: Remove `()`

sourced/ml/algorithms/uast2nodes.py
```diff
@@ -0,0 +1,131 @@
...
+        """
+        :param uast: The UAST root node.
+        :return: The list of the nodes composing the UAST.
+        """
+        nodes = []
+        queue = [root]
+        while queue:
+            child = queue.pop(0)
+            queue.extend(child.children)
+            nodes.append(child)
```
vmarkovtsev: Better `yield child` and delete `nodes`

sourced/ml/algorithms/uast2nodes.py
```diff
@@ -0,0 +1,131 @@
...
+        The number of children are preprocessed by :class:`Quantization`
+
+        :param uast: The UAST root node.
+        :param feature: The first feature we want to extract from the nodes of the UAST.
+        :return:
+        """
+        bag = defaultdict(int)
+        nodes = self._uast2nodes(uast)
+        nb_children = self._nodes2children(nodes)
+        self.children2ind.build_partition(nb_children)
```
vmarkovtsev: We cannot do this at this stage: indices will be different for every file.

You need to build the global partition during the docfreq stage and assign it.

sourced/ml/algorithms/uast2nodes.py
vmarkovtsev: We should erase "roles" and delete all the `feature` code - we don't need it anyway.

sourced/ml/algorithms/uast2nodes.py
```diff
@@ -0,0 +1,131 @@
...
+        """
+        str format is required for wmhash.Bags.Extractor
+        :param node: a node of UAST
+        :param feature: The first feature we want to extract from the nodes of the UAST. \
+        This first feature should be either "internal_type" or "roles", \
+        the second being the number of children anyway.
+        :return:
+        """
+        if feature is "internal_type":
+            return " ".join([self.type2ind[node.internal_type],
```
vmarkovtsev: No need for `type2ind`, remove it and dump raw strings.

vmarkovtsev: `" "` is a bad value, better `"_"`

sourced/ml/algorithms/uast2nodes.py
```diff
@@ -0,0 +1,131 @@
...
+                             str(self.children2ind.process_value(len(node.children)))])
+        elif feature is "roles":
+            return " ".join([str(self.roles2ind[node.roles]),
+                             str(self.children2ind.process_value(len(node.children)))])
+        else:
+            raise NotImplementedError
+
+
+class Quantization():
```
vmarkovtsev: `()`

sourced/ml/algorithms/uast2nodes.py
```diff
@@ -0,0 +1,131 @@
...
+    """
+    Algorithm that performs the quantization of a list of values.
+    """
+    def __init__(self, fineness=0.005):
```
vmarkovtsev: It is not convenient to set quantization like this. We should accept the target number of partitions instead.

sourced/ml/algorithms/uast2nodes.py
```diff
@@ -0,0 +1,131 @@
...
+        Builds the partition of the quantization.
+        It is a list of increasing integers equally partitioning the distribution of values.
+        The length of the partition list increase with the fineness.
+
+        :param values: List of values we want to quantize
+        :return:
+        """
+        values.sort()
+        max_nodes_per_bin = self.fineness * len(values)
+        unique_values = list(set(values))
```
vmarkovtsev: `set(values)` is unordered yet down the code you have `unique_values[ind]`

sourced/ml/algorithms/uast2nodes.py
warenlg: I don't get "dump raw strings" as join() requires a sequence of strings and int is not supported

vmarkovtsev:
@warenlg Please create the new PR because I cannot reopen this one - `refactor` was deleted.

PR #144, opened 2017-12-19
Move cmd entries to submodule

zurk:
As was discussed on refactoring review I create new submodule and move main command line entry points there. 

I decide not to move `bigartm` installation entry, leave it in utils.
I decide not to move all entries for id_embeddings except `postprocess` and `run_swivel` since all others should be mainly refactored.


sourced/ml/__main__.py
```diff
@@ -4,12 +4,10 @@
 import sys
 
 from modelforge.logs import setup_logging
-from sourced.ml.algorithms.id_embedding import preprocess as preprocess_id2vec, run_swivel, \
-    postprocess as postprocess_id2vec, swivel
-from sourced.ml.models.bow import bow2vw_entry
-from sourced.ml.utils import install_bigartm, bigartm2asdf_entry
-from sourced.ml.utils import dump_model
-from sourced.ml.utils import projector_entry
+from sourced.ml.algorithms.id_embedding import preprocess as preprocess_id2vec, swivel
```
vmarkovtsev: there can be no imports from `algorithms` here. Move/split.

zurk: But I explain in a comment why it is here:
> I decide not to move all entries for id_embeddings except postprocess and run_swivel since all others should be mainly refactored.

Do you not agree?

vmarkovtsev: Just extract this function into a separate file inside `cmd_entries`, no need to do any refactoring

zurk: ok

PR #145, opened 2017-12-19
Uast2all refactoring

zurk:
Here I propose my point of view how uast2bow-type algorithm should look like.
Some points were already discussed, some of them not. 

@EgorBu, this refactoring affects a lot of your code, I try to make it better from my point of view, so feel free not to agree with me. Let's discuss. 

I add comments to lines in this commit to make it clearer what I suggest to change and why.

### DIFF
--- a/sourced/ml/algorithms/token_parser.py
+++ b/sourced/ml/algorithms/token_parser.py
@@ -17,6 +17,9 @@ class TokenParser:
         self._stem_threshold = stem_threshold
         self._max_token_length = max_token_length
 
+    def __call__(self, token):
+        return self.process_token(token)
+
     def process_token(self, token):
         for word in self.split(token):
             yield self.stem(word)
@@ -79,3 +82,6 @@ class NoopTokenParser:
 
     def process_token(self, token):
         yield token
+
+    def __call__(self, token):
+        return self.process_token(token)
--- a/sourced/ml/algorithms/uast_ids_to_bag.py
+++ b/sourced/ml/algorithms/uast_ids_to_bag.py
@@ -56,7 +56,7 @@ class UastTokens2Bag(Uast2BagBase):
         nodes = bblfsh.filter(uast, self.XPATH)
         bag = defaultdict(int)
         for node in nodes:
-            for sub in self._token_parser(node.token):
+            for sub in self._token_parser.process_token(node.token):
                 try:
                     bag[self._token2index[sub]] += 1
                 except KeyError:
--- a/sourced/ml/algorithms/uast_ids_to_bag.py
+++ b/sourced/ml/algorithms/uast_ids_to_bag.py
@@ -1,6 +1,8 @@
 from collections import defaultdict
 
-from sourced.ml.algorithms import TokenParser
+import bblfsh
+
+from sourced.ml.algorithms import TokenParser, NoopTokenParser
 
 
 class FakeVocabulary:
@@ -8,39 +10,72 @@ class FakeVocabulary:
         return item
 
 
-class UastIds2Bag:
+class Uast2BagBase:
     """
-    Converts a UAST to a bag-of-identifiers.
+    Base class to convert UAST to bag of anything.
     """
-    def __init__(self, vocabulary, token_parser=None):
+    def __call__(self, uast):
         """
-        :param vocabulary: The mapping from tokens to bag keys. If None, no mapping is performed.
-        :param token_parser: Specify token parser if you want to use a custome one. \
-            :class:'TokenParser' is used if it is not specified.
-        """
-        self._vocabulary = FakeVocabulary() if vocabulary is None else vocabulary
-        self._token_parser = TokenParser() if token_parser is None else token_parser
-
-    @property
-    def vocabulary(self):
-        return self._vocabulary
-
-    def uast_to_bag(self, uast, roles_filter="//*[@roleIdentifier and not(@roleQualified)]"):
-        """
-        Converts a UAST to a bag-of-words. The weights are identifier frequencies.
-        The identifiers are preprocessed by :class:`TokenParser`.
+        Override this function with converting algorithm.
 
         :param uast: The UAST root node.
-        :param roles_filter: The libuast xpath query to filter identifiers.
+        :return: bag
+        """
+        raise NotImplemented
+
+
+class UastTokens2Bag(Uast2BagBase):
+    """
+    Converts a UAST to a weighed bag of tokens via xpath.
+    """
+
+    XPATH = None  # Should be overridden in child class
+
+    def __init__(self, token2index=None, token_parser=None):
+        """
+        :param xpath: The libuast xpath query to filter nodes with needed tokens.
+        :param token2index: The mapping from tokens to bag keys. If None, no mapping is performed.
+        :param token_parser: Specify token parser if you want to use a custom one. \
+            :class:'NoopTokenParser' is used if it is not specified.
+        """
+        self._token2index = FakeVocabulary() if token2index is None else token2index
+        self._token_parser = NoopTokenParser() if token_parser is None else token_parser
+
+    @property
+    def token2index(self):
+        return self._token2index
+
+    def __call__(self, uast):
+        """
+        Converts a UAST to a weighed bag-of-words. The weights are words frequencies.
+        The tokens are preprocessed by _token_parser.
+
+        :param uast: The UAST root node.
         :return:
         """
-        import bblfsh
-        nodes = bblfsh.filter(uast, roles_filter)
+        nodes = bblfsh.filter(uast, self.XPATH)
         bag = defaultdict(int)
         for node in nodes:
-            for sub in self._token_parser.process_token(node.token):
+            for sub in self._token_parser(node.token):
                 try:
-                    bag[self._vocabulary[sub]] += 1
+                    bag[self._token2index[sub]] += 1
                 except KeyError:
                     continue
         return bag
+
+
+class UastIds2Bag(UastTokens2Bag):
+    """
+    Converts a UAST to a bag-of-identifiers.
+    """
+
+    XPATH = "//*[@roleIdentifier and not(@roleQualified)]"
+
+    def __init__(self, token2index=None, token_parser=None):
+        """
+        :param token2index: The mapping from tokens to bag keys. If None, no mapping is performed.
+        :param token_parser: Specify token parser if you want to use a custom one. \
+            :class:'TokenParser' is used if it is not specified.
+        """
+        token_parser = TokenParser() if token_parser is None else token_parser
+        super().__init__(token2index, token_parser)
--- a/sourced/ml/algorithms/uast_struct_to_bag.py
+++ b/sourced/ml/algorithms/uast_struct_to_bag.py
@@ -17,6 +17,11 @@ class Uast2StructBagBase(Uast2BagBase):
         return self._node2index
 
 
+class Node2InternalType:
+    def __getitem__(self, item):
+        return item.internal_type
+
+
 class UastSeq2Bag(Uast2StructBagBase):
     """
     DFS traversal + preserves the order of node children.
@@ -67,11 +72,6 @@ class Node:
         return neighbours
 
 
-class Node2InternalType:
-    def __getitem__(self, item):
-        return item.internal_type
-
-
 class Uast2RandomWalks:
     """
     Generation of random walks for UAST.
@@ -80,6 +80,8 @@ class Uast2RandomWalks:
     def __init__(self, p_explore_neighborhood, q_leave_neighborhood, n_walks, n_steps,
                  node2index=None, seed=None):
         """
+        Related article: https://arxiv.org/abs/1607.00653
+
         :param p_explore_neighborhood: return parameter, p. Parameter p controls the likelihood of\
                                        immediately revisiting a node in the walk. Setting it to a\
                                        high value (> max(q, 1)) ensures that we are less likely to\
@@ -139,10 +141,12 @@ class Uast2RandomWalks:
         """
         Compare to node2vec this sampling is a bit simpler because there is no loop in tree ->
         so there are only 2 options with unnormalized probabilities 1/p & 1/q
+        Related article: https://arxiv.org/abs/1607.00653
+
         :param walk: list of visited nodes
         :return: next node to visit
         """
-        last_node = walk[-1]
+        last_node = walk[-1]  # correspond to node v in article
 
         if len(walk) == 1 and len(last_node.children) > 0:
             return random.choice(last_node.children)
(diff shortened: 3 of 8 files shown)


sourced/ml/algorithms/uast_ids_to_bag.py
```diff
@@ -1,46 +1,81 @@
...
+        :param token_parser: Specify token parser if you want to use a custom one. \
+            :class:'NoopTokenParser' is used if it is not specified.
+        """
+        self._token2index = FakeVocabulary() if token2index is None else token2index
+        self._token_parser = NoopTokenParser() if token_parser is None else token_parser
 
     @property
-    def vocabulary(self):
-        return self._vocabulary
+    def token2index(self):
```
zurk: hard to understand what is `vocabulary` means. Easier to get it if a name is `token2index`.

sourced/ml/algorithms/uast_ids_to_bag.py
```diff
@@ -1,46 +1,81 @@
...
+
+    XPATH = "//*[@roleIdentifier and not(@roleQualified)]"
+
+    def __init__(self, token2index=None, token_parser=None):
+        """
+        :param token2index: The mapping from tokens to bag keys. If None, no mapping is performed.
+        :param token_parser: Specify token parser if you want to use a custom one. \
+            :class:'TokenParser' is used if it is not specified.
+        """
+        token_parser = TokenParser() if token_parser is None else token_parser
```
zurk: override default `token_parser` for identifiers.

sourced/ml/algorithms/uast_struct_to_bag.py
```diff
@@ -1,24 +1,32 @@
...
 from collections import defaultdict, deque
 
-from sourced.ml.algorithms.uast_ids_to_bag import FakeVocabulary
+from sourced.ml.algorithms.uast_ids_to_bag import FakeVocabulary, Uast2BagBase
 
 
-class UastStructure2BagBase:
-    def uast_to_bag(self, uast):
-        raise NotImplemented
+class Uast2StructBagBase(Uast2BagBase):
```
zurk: base class for all structural bags extractions

sourced/ml/algorithms/uast_struct_to_bag.py
```diff
@@ -1,24 +1,32 @@
...
+    def __init__(self, stride, seq_len, node2index=None):
+        self._node2index = node2index if node2index is not None else FakeVocabulary()
+        self._stride = stride
+        if not isinstance(seq_len, (int, tuple, list)):
+            raise TypeError("Unexpected type of seq_len: %s" % type(seq_len))
+        self._seq_lens = [seq_len] if isinstance(seq_len, int) else seq_len
+
+    @property
+    def node2index(self):
```
zurk: similar to `token2index` property. But here we do not have `token_parser` stage.

sourced/ml/algorithms/uast_struct_to_bag.py
```diff
@@ -1,24 +1,32 @@
...
+class UastSeq2Bag(Uast2StructBagBase):
     """
     DFS traversal + preserves the order of node children.
     """
-    def __init__(self, type2ind=None, stride=1, seq_len=5):
-        self.type2ind = type2ind if type2ind is not None else FakeVocabulary()
-        self.stride = stride
-        self.seq_len = seq_len
+    def __init__(self, stride=1, seq_len=5, node2index=None):
+        _node2index = Node2InternalType() if node2index is None else node2index
```
zurk: `Node2InternalType ` definition is below (https://github.com/src-d/ml/pull/145/files#diff-86f78a3692ae357bb5a7628fcf0cbc6bR70). Think that I need to lift it up before this class.

sourced/ml/algorithms/uast_struct_to_bag.py
```diff
@@ -1,24 +1,32 @@
...
+        super().__init__(stride, seq_len, _node2index)
 
-    def _uast2seq(self, root, walk):
+    def _uast2sequence(self, root):
```
zurk: full names are always better. Also, I return walk directly from this function.

sourced/ml/algorithms/uast_struct_to_bag.py
```diff
@@ -28,44 +36,49 @@ def _uast2seq(self, root, walk):
...
         bag = defaultdict(int)
-        seq = []
-        self._uast2seq(uast, seq)
+        node_sequence = self._uast2sequence(uast)
 
         # convert to str - requirement from wmhash.BagsExtractor
-        seq = "".join([self.node2ind(n) for n in seq])
+        node_sequence = "".join([self.node2index[n] for n in node_sequence])
 
-        if isinstance(self.seq_len, int):
```
zurk: this part was moved here https://github.com/src-d/ml/pull/145/files#diff-86f78a3692ae357bb5a7628fcf0cbc6bR11 up to a parent class.

sourced/ml/algorithms/uast_struct_to_bag.py
```diff
@@ -28,44 +36,49 @@ def _uast2seq(self, root, walk):
...
-
-        for seq_len in seq_lens:
-            for i in range(0, len(seq) - seq_len + 1, self.stride):
-                bag[seq[i:i + seq_len]] += 1
+        for seq_len in self._seq_lens:
+            for i in range(0, len(node_sequence) - seq_len + 1, self._stride):
+                bag[node_sequence[i:i + seq_len]] += 1
         return bag
 
-    def node2ind(self, node):
```
zurk: now we have `_node2index` dict-like object. As was earlier coded in `UastIds2Bag`. 
Want to make it consistent.

sourced/ml/algorithms/uast_struct_to_bag.py
```diff
@@ -87,103 +100,77 @@ def __init__(self, p_explore_neighborhood, q_leave_neighborhood, n_walks, n_step
         self.q_leave_neighborhood = q_leave_neighborhood
         self.n_walks = n_walks
         self.n_steps = n_steps
-        self.type2ind = type2ind if type2ind is not None else FakeVocabulary()
+        self.node2index = node2index if node2index is not None else Node2InternalType()
+        if seed is not None:
```
zurk: Move seed here from `UastRandomWalk2Bag ` class because here all random parts.

Note, that If earlier it was inside `uast_to_bag` function, now it is in `__init__`. And I think it is better.

sourced/ml/algorithms/uast_struct_to_bag.py
```diff
@@ -87,103 +100,77 @@ def __init__(self, p_explore_neighborhood, q_leave_neighborhood, n_walks, n_step
...
 
-    def uast2walks(self, uast):
+    def __call__(self, uast):
         starting_nodes = self.prepare_starting_nodes(uast)
-        res = []
         for i in range(self.n_walks):
             for start_node in starting_nodes:
-                res.append(self.random_walk(start_node))
-        return res
+                yield self.random_walk(start_node)
```
zurk: An iterator is better due to next `__call__ ` usage here https://github.com/src-d/ml/pull/145/files#diff-86f78a3692ae357bb5a7628fcf0cbc6bR171

sourced/ml/algorithms/uast_struct_to_bag.py
```diff
@@ -87,103 +100,77 @@ def __init__(self, p_explore_neighborhood, q_leave_neighborhood, n_walks, n_step
...
 
     @staticmethod
     def _extract_node(node, parent):
         return Node(parent=parent, internal_type=node.internal_type)
 
     def prepare_starting_nodes(self, uast):
         starting_nodes = []
-        self._prepare_starting_nodes(uast, None, starting_nodes)
```
zurk: use stack instead of recursion. I check the result with equivalence test. It gives the same result. I can add it to tests If you wish.

sourced/ml/algorithms/uast_struct_to_bag.py
```diff
@@ -87,103 +100,77 @@ def __init__(self, p_explore_neighborhood, q_leave_neighborhood, n_walks, n_step
...
     def alias_sample(self, walk):
         """
         Compare to node2vec this sampling is a bit simpler because there is no loop in tree ->
         so there are only 2 options with unnormalized probabilities 1/p & 1/q
         :param walk: list of visited nodes
         :return: next node to visit
         """
-        # notation from article - t - node from previous step, v - last node
-        v = walk[-1]
+        last_node = walk[-1]
```
zurk: https://github.com/src-d/guide/blob/master/engineering/python-style-guide.md
> Do not use single letter argument names; use X and Y only in sklearn context.

sourced/ml/algorithms/uast_struct_to_bag.py
```diff
@@ -87,103 +100,77 @@ def __init__(self, p_explore_neighborhood, q_leave_neighborhood, n_walks, n_step
...
         threshold = (1 / self.p_explore_neighborhood)
-        threshold /= ((1 / self.p_explore_neighborhood) +
-                      len(v.children) / self.q_leave_neighborhood)
+        threshold /= (threshold + len(last_node.children) / self.q_leave_neighborhood)
 
         if random.random() <= threshold:
-            return t
+            return walk[-2]
 
-        options = []
```
zurk: move this part to `Node` class, because it is only about `Node`.

sourced/ml/algorithms/uast_struct_to_bag.py
```diff
@@ -87,103 +100,77 @@ def __init__(self, p_explore_neighborhood, q_leave_neighborhood, n_walks, n_step
...
-            random.seed(self.seed)
+        super().__init__(stride, seq_len)
+        self.uast2walks = Uast2RandomWalks(p_explore_neighborhood=p_explore_neighborhood,
+                                           q_leave_neighborhood=q_leave_neighborhood,
+                                           n_walks=n_walks, n_steps=n_steps, seed=seed)
 
+    def __call__(self, uast):
         bag = defaultdict(int)
-        walks = self.random_walker.uast2walks(uast)
-        if isinstance(self.seq_len, int):
```
zurk: it is in the `__init__` of base class

sourced/ml/algorithms/uast_struct_to_bag.py
EgorBu: if not one character - please make comment with mapping to variable names in the article.

sourced/ml/algorithms/uast_struct_to_bag.py
```diff
@@ -87,103 +100,77 @@ def __init__(self, p_explore_neighborhood, q_leave_neighborhood, n_walks, n_step
...
-            seq_lens = [self.seq_len]
-        else:
-            seq_lens = self.seq_len
-        for walk in walks:
-            for seq_len in seq_lens:
-                for i in range(0, len(walk) - seq_len, self.stride):
+        for walk in self.uast2walks(uast):
+            for seq_len in self._seq_lens:
+                for i in range(0, len(walk) - seq_len, self._stride):
```
EgorBu: `len(walk) - seq_len` -> `len(walk) - seq_len + 1` not to lose last element - bug in initial implementation

sourced/ml/algorithms/uast_struct_to_bag.py
```diff
@@ -87,103 +100,77 @@ def __init__(self, p_explore_neighborhood, q_leave_neighborhood, n_walks, n_step
...
-        return node
+        return starting_nodes
 
     def random_walk(self, node):
         walk = [node]
         while len(walk) < self.n_steps:
             walk.append(self.alias_sample(walk))
 
-        walk = [self.node2feat(n) for n in walk]
+        walk = [self.node2index[n] for n in walk]
```
EgorBu: comment above about node2index

sourced/ml/algorithms/uast_struct_to_bag.py
zurk: but you can do the same in `__getitem__` style. 
also, it will allow you to split this algorithm from "find most similar in dictionary" one.
Also if you need any state to find the similar combination of roles (why not) it is better to have a class.

sourced/ml/algorithms/uast_struct_to_bag.py
zurk: ok!

sourced/ml/algorithms/uast_struct_to_bag.py
```diff
@@ -87,103 +100,77 @@ def __init__(self, p_explore_neighborhood, q_leave_neighborhood, n_walks, n_step
...
+            return last_node
 
-        t = walk[-2]
         threshold = (1 / self.p_explore_neighborhood)
-        threshold /= ((1 / self.p_explore_neighborhood) +
-                      len(v.children) / self.q_leave_neighborhood)
+        threshold /= (threshold + len(last_node.children) / self.q_leave_neighborhood)
 
         if random.random() <= threshold:
-            return t
```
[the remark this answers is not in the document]
zurk: make sense. 
But we need to create `t` (or whatever variable) just to return it.

What do you think if I just add a comment to `return walk[-2]` with explanation what is happening?

sourced/ml/algorithms/uast_struct_to_bag.py
EgorBu: make sense

sourced/ml/algorithms/uast_struct_to_bag.py
EgorBu: tests are always welcome)

sourced/ml/algorithms/uast_struct_to_bag.py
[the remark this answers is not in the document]
EgorBu: the explanation must be provided. Name of variable may be omitted

sourced/ml/algorithms/token_parser.py
```diff
@@ -17,6 +17,9 @@ def __init__(self, stem_threshold=STEM_THRESHOLD, max_token_length=MAX_TOKEN_LEN
         self._stem_threshold = stem_threshold
         self._max_token_length = max_token_length
 
+    def __call__(self, token):
```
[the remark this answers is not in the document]
vmarkovtsev: ParseToken is a really bad name

sourced/ml/algorithms/uast_struct_to_bag.py
```diff
@@ -28,45 +41,47 @@ def _uast2seq(self, root, walk):
...
+        node_sequence = self._uast2sequence(uast)
 
         # convert to str - requirement from wmhash.BagsExtractor
-        seq = "".join([self.node2ind(n) for n in seq])
-
-        if isinstance(self.seq_len, int):
-            seq_lens = [self.seq_len]
-        else:
-            seq_lens = self.seq_len
+        node_sequence = "".join([self.node2index[n] for n in node_sequence])
```
vmarkovtsev: No need to have a list, a plain generator is also ok

PR #149, opened 2017-12-22
Cacher `maybe` function refactor

zurk:
Now you can do the following:
```python
...
.link(Cacher.maybe(args.persist))
...
```

And before it was not consistent. Check in https://github.com/src-d/ml/pull/149/files#diff-653487ec5c95d3a894f70df54a56a7ac

### DIFF
--- a/sourced/ml/transformers/basic.py
+++ b/sourced/ml/transformers/basic.py
@@ -15,6 +15,11 @@ class First(Transformer):
         return head.first()
 
 
+class Identity(Transformer):
+    def __call__(self, head):
+        return head
+
+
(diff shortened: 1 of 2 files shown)


sourced/ml/transformers/basic.py
```diff
@@ -15,6 +15,11 @@ def __call__(self, head):
         return head.first()
 
 
+class Equal(Transformer):
```
vmarkovtsev: "Identity"  is better

zurk: done

PR #155, opened 2017-12-27
Fix memory config + initialization of transformers

EgorBu:
Fix initialization of transformers:
- add **kwargs where it's missed
-  pass **kwargs to a parent class

Fix passing memory config to engine + make error logs

### DIFF
--- a/sourced/ml/utils/engine.py
+++ b/sourced/ml/utils/engine.py
@@ -3,7 +3,7 @@ import os
 import sys
 os.environ["PYSPARK_PYTHON"] = sys.executable
 
-import pyspark
+import pyspark  # nopep8
 from pyspark.sql import SparkSession  # nopep8
 from sourced.engine import Engine  # nopep8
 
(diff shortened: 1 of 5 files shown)


sourced/ml/utils/engine.py
```diff
@@ -105,12 +105,16 @@ def create_engine(session_name, repositories,
...
+    memory_conf = []
     if memory:
         memory = memory.split(",")
-        memory.append("spark.executor.memory=" + memory[0])
-        memory.append("spark.driver.memory=" + memory[1])
-        memory.append("spark.driver.maxResultSize=" + memory[2])
-    session = create_spark(session_name, config=config, packages=packages, memory=memory,
+        err = "Expected 3 memory parameters but got {}. " \
+              "Please check --help to see how -m/--memory should be used."
+        assert len(memory) == 3, err.format(len(memory))
```
zurk: I think it is better to use
```
if len(memory) == 3:
    raise ValueError(err)
```

`assert ` more for debug purposes.

EgorBu: `raise` can be useful if you want to handle it somehow. I don't think wrong parameters to script should be handled in this situation because it may lead to some unexpected behaviour for future users. And you right - the purpose was to make it easy to understand the error by users.

PR #159, opened 2018-01-11
Save the quantization mapping in the docfreq model

warenlg:


sourced/ml/models/df.py
```diff
@@ -12,12 +12,14 @@ class DocumentFrequencies(Model):
     """
     NAME = "docfreq"
 
-    def construct(self, docs: int, tokfreqs: Union[Iterable[Dict[str, int]], Dict[str, int]]):
+    def construct(self, docs: int, tokfreqs: Union[Iterable[Dict[str, int]],
+                  Dict[str, int]], quant_map: Dict[str, int]=None):
```
vmarkovtsev: No need to extend this interface, pls leave as is

The key prefix allows to tell the identity

sourced/ml/transformers/batch_transformers.py
```diff
@@ -38,8 +38,14 @@ def __call__(self, processed):
...
         self._log.info("Number of documents: %d", ndocs)
+        quant_map = None
+        for e in self.extractors:
+            try:
+                quant_map = e.quant_map
+            except AttributeError:
+                continue
         self.model = OrderedDocumentFrequencies().construct(
-            self.extractors[0].ndocs, [e.docfreq for e in self.extractors])
+            self.extractors[0].ndocs, [e.docfreq for e in self.extractors], quant_map)
```
vmarkovtsev:
```
[e.docfreq for e in self.extractors] + [quant_map])
```

sourced/ml/models/df.py
```diff
@@ -24,8 +24,16 @@ def construct(self, docs: int, tokfreqs: Union[Iterable[Dict[str, int]], Dict[st
             df = tokfreqs
         else:
             df = {}
-            for d in tokfreqs:
-                df.update(d)
+            self.quant_map = tokfreqs[-1]
```
vmarkovtsev: Why do you need to catch `TypeError`? We should not change this model at all.

sourced/ml/transformers/batch_transformers.py
```diff
@@ -38,8 +38,14 @@ def __call__(self, processed):
...
         self._log.info("Number of documents: %d", ndocs)
+        quant_map = None
+        for e in self.extractors:
+            try:
+                quant_map = e.quant_map
+            except AttributeError:
+                continue
         self.model = OrderedDocumentFrequencies().construct(
-            self.extractors[0].ndocs, [e.docfreq for e in self.extractors])
+            self.extractors[0].ndocs, [e.docfreq for e in self.extractors] + [quant_map])
```
vmarkovtsev: Now I see. Well, just do this

```
dicts = [e.docfreq for e in self.extractors]
for e in self.extractors:
    try:
        dicts.append(e.quant_map)
    except AttributeError:
        continue
```

sourced/ml/models/df.py
warenlg: I don't get you, if you don't want me to touch this model, quant_map will be stored in self._df

vmarkovtsev: Exactly

PR #160, opened 2018-01-11
Add repo2bow entry point

zurk:
I add `repo2bow` entry point. But for now, I do not like how it looks.

What I was needed to change:
1. Process with `Repo2WeightedSet` (renamed to `Documents2BOW `) not only separate files but repositories as well. And also any other structure that we can use as a document.
2. The same for `Repo2DocFreq ` (renamed to `Documents2DocFreq `)

It is the easiest way to implement `repo2bow` entry point, but think that not too flexible. 
I still don't like tight contention between `Documents2BOW` and `Documents2DocFreq` and df processing hidden inside `Documents2DocFreq` and `extractors`. Maybe we can change it somehow?

For now out of scope `bow` model refactoring. 
Need to delete unnecessary code and rename something. 

P.S.: PR is not polished and need for discussion. It is not ready to merge.


sourced/ml/cmd_entries/repos2bow.py
```diff
@@ -0,0 +1,52 @@
...
+    document_column_name = "repository_id"
+    extractors = [__extractors__[s](
+        args.min_docfreq, **__extractors__[s].get_kwargs_fromcmdline(args))
+        for s in args.feature]
+
+    uast = Engine(engine, explain=args.explain) \
+        .link(HeadFiles()) \
+        .link(UastExtractor(languages=args.languages)) \
+        .link(UastDeserializer())
+    df = uast.link(Documents2DocFreq(extractors, document_column_name))
```
vmarkovtsev: We can add Waren's quantizer here too

sourced/ml/cmd_entries/repos2bow.py
```diff
@@ -0,0 +1,52 @@
...
+    doc_bow = res.take(res.count())
+
+    tokens = []
+    documents = []
+    documents2id = {}
+    for doc, bow in doc_bow:
+        tokens += bow.keys()
+        documents.append(doc)
+        documents2id[doc] = len(documents2id)
+    tokens = list(set(tokens))
```
vmarkovtsev: Need to sort otherwise the order is different every time

sourced/ml/cmd_entries/repos2bow.py
```diff
@@ -0,0 +1,52 @@
...
+    tokens2id = {t: i for i, t in enumerate(tokens)}
+
+    matrix = sparse.csc_matrix((len(documents), len(tokens)))
+    for doc, bow in doc_bow:
+        for token in bow:
+            matrix[documents2id[doc], tokens2id[token]] = bow[token]
+
+    model_df = OrderedDocumentFrequencies() if args.ordered else DocumentFrequencies()
```
vmarkovtsev: would be nice to move upper `tokens = []`

sourced/ml/transformers/repo2df.py
```diff
@@ -1,33 +1,30 @@
 from sourced.ml.transformers import Transformer
 
 
-class Repo2DocFreq(Transformer):
-    NDOCS_KEY = -1, 0
+class Documents2DocFreq(Transformer):
 
-    def __init__(self, extractors, **kwargs):
+    def __init__(self, extractors, document_column='blob_id', **kwargs):
```
vmarkovtsev: We should not give this (or any) default

sourced/ml/transformers/repo2df.py
```diff
@@ -1,33 +1,30 @@
...
         for extractor in self.extractors:
             extractor.ndocs = ndocs
 
-    def process_row(self, row):
-        yield self.NDOCS_KEY, 1
+    def _process_row(self, row):
         for i, extractor in enumerate(self.extractors):
             for k in extractor.inspect(row.uast):
-                yield (i, k), 1
+                yield i, k, row.asDict()[self.document_column]
```
vmarkovtsev: `getattr(row, self.document_column)` is faster

sourced/ml/transformers/repo2df.py
vmarkovtsev: And according to the default value, the name is bad. Should be `id_column`

sourced/ml/transformers/repo2df.py
```diff
@@ -1,33 +1,30 @@
...
+        processed = rows.flatMap(self._process_row)
         if self.explained:
             self._log.info("toDebugString():\n%s", processed.toDebugString().decode())
-        reduced = processed.countByKey()
-        ndocs = None
+        ndocs = rows.map(lambda x: x.asDict(self.document_column)).count()
+        reduced = processed.distinct().map(lambda x: ((x[0], x[1]), 1)).countByKey()
+
         for (i, key), value in reduced.items():
-            if (i, key) == self.NDOCS_KEY:
```
vmarkovtsev: This is not efficient. The new code does two passes while the old code does one. Please revert

sourced/ml/transformers/repo2df.py
```diff
@@ -1,33 +1,30 @@
...
 
     def __call__(self, rows):
-        processed = rows.flatMap(self.process_row)
+        processed = rows.flatMap(self._process_row)
         if self.explained:
             self._log.info("toDebugString():\n%s", processed.toDebugString().decode())
-        reduced = processed.countByKey()
-        ndocs = None
+        ndocs = rows.map(lambda x: x.asDict(self.document_column)).count()
+        reduced = processed.distinct().map(lambda x: ((x[0], x[1]), 1)).countByKey()
```
vmarkovtsev: `distinct` should not be called. It is the responsibility of the caller

sourced/ml/transformers/repo2weighted.py
```diff
@@ -1,16 +1,28 @@
+import operator
+
 from sourced.ml.transformers import Transformer
 
 
-class Repo2WeightedSet(Transformer):
-    def __init__(self, extractors, **kwargs):
+class Documents2BOW(Transformer):
+    def __init__(self, extractors, document_column="blob_id", **kwargs):
```
vmarkovtsev: Again, no default

sourced/ml/transformers/repo2weighted.py
```diff
@@ -1,16 +1,28 @@
+import operator
+
 from sourced.ml.transformers import Transformer
 
 
-class Repo2WeightedSet(Transformer):
-    def __init__(self, extractors, **kwargs):
+class Documents2BOW(Transformer):
```
vmarkovtsev: Why not weighted? It is weighted

sourced/ml/transformers/repo2df.py
```diff
@@ -1,33 +1,30 @@
...
 
         for extractor in self.extractors:
             extractor.ndocs = ndocs
 
-    def process_row(self, row):
-        yield self.NDOCS_KEY, 1
+    def _process_row(self, row):
         for i, extractor in enumerate(self.extractors):
             for k in extractor.inspect(row.uast):
-                yield (i, k), 1
```
vmarkovtsev: This breaks the upstream logic which relies on the composite key

sourced/ml/transformers/repo2weighted.py
```diff
@@ -1,16 +1,28 @@
...
+    def __init__(self, extractors, document_column="blob_id", **kwargs):
         super().__init__(**kwargs)
         self.extractors = extractors
+        self.document_column = document_column
 
     def __call__(self, rows):
-        return rows.map(self.process_row)
```
vmarkovtsev: The code was so simple, so innocent. Why did you touch it at all?

sourced/ml/cmd_entries/repos2bow.py
zurk: you mean like an option?

sourced/ml/transformers/repo2df.py
zurk: default just for back compatibility. will be removed. Just need to make changes in Apollo also.

`id_column ` also bad. I can read it as `identifiers column`. can rename to `document_id`?

sourced/ml/transformers/repo2df.py
zurk: I need. `distinct()` here to count everything right. 
All this class has a tight connection to work with only one uast, so I need to do some tricks to make it work for any "document". Also that the reason why I can not use this special `NDOCS_KEY` to count documents anymore.

sourced/ml/transformers/repo2df.py
zurk: Commented above.

sourced/ml/transformers/repo2df.py
zurk: Not really, also that the reason for renaming. This method is used only in `__call__` method of this class.

sourced/ml/cmd_entries/repos2bow.py
vmarkovtsev: Obligatory, some extractors require an additional "quantization" step

sourced/ml/transformers/repo2weighted.py
zurk: We discuss it when I was in Madrid. BOW = Weighted set. So just to avoid different namings for the same thing. BOW was the first.

sourced/ml/transformers/repo2weighted.py
zurk: because I need to merge all document parts from different rows together.

sourced/ml/transformers/repo2df.py
vmarkovtsev: We can but this file has nothing to do with identifiers.
`document_id` does not imply that we expect a column name.
We can use `key_column`

sourced/ml/transformers/repo2weighted.py
vmarkovtsev: We urgently need a glossary of terms like I have in Apollo https://github.com/src-d/apollo/blob/master/doc/GLOSSARY.md

zurk:
About the last commit.
PR still in progress, and this commit surely breaks something, but I just need to be sure that I am going in the right direction.

**Main change**: Now we calculate TFIDF in another Transformer from separately collected TF and DF. BagExtractor collects pure TF.

If everything ok, I will continue to adopt all other code to use this scheme of TF-IDF calculating, add comments to the new code, and think that tests also. After that we can merge this PR.

sourced/ml/cmd_entries/repos2bow.py
```diff
@@ -0,0 +1,45 @@
...
+
+    uast_pipeline = Engine(engine, explain=args.explain) \
+        .link(HeadFiles()) \
+        .link(UastExtractor(languages=args.languages)) \
+        .link(UastDeserializer())
+    df_pipeline = uast_pipeline.link(Uast2DocFreq(extractors, document_column_name))
+    df = df_pipeline.execute()
+    tf_pipeline = uast_pipeline.link(Uast2TermFreq(extractors, document_column_name))
+    tf = tf_pipeline.execute()
+    tfidf = TFIDF(df, tf).execute()
```
vmarkovtsev: :+1:

sourced/ml/cmd_entries/repos2bow.py
```diff
@@ -0,0 +1,45 @@
...
+
+    tfidf = tfidf.take(tfidf.count())
```
vmarkovtsev: WAT

sourced/ml/transformers/repo2df.py
```diff
@@ -1,33 +1,31 @@
...
+
 from sourced.ml.transformers import Transformer
 
 
-class Repo2DocFreq(Transformer):
-    NDOCS_KEY = -1, 0
+class Uast2DocFreq(Transformer):
 
-    def __init__(self, extractors, **kwargs):
+    def __init__(self, extractors, document_column='blob_id', **kwargs):
```
vmarkovtsev: Remove the defaults

sourced/ml/transformers/repo2df.py
```diff
@@ -1,33 +1,31 @@
...
-        ndocs = None
-        for (i, key), value in reduced.items():
-            if (i, key) == self.NDOCS_KEY:
-                ndocs = value
-                continue
-            self.extractors[i].apply_docfreq(key, value)
 
-        for extractor in self.extractors:
-            extractor.ndocs = ndocs
+        df = processed \
```
vmarkovtsev: All good, just please make it a single "return" expression to be shorter (include upper line too)

sourced/ml/transformers/repo2weighted.py
```diff
@@ -1,16 +1,23 @@
+import operator
+
 from sourced.ml.transformers import Transformer
 
 
-class Repo2WeightedSet(Transformer):
-    def __init__(self, extractors, **kwargs):
+class Uast2TermFreq(Transformer):
+    def __init__(self, extractors, document_column="blob_id", **kwargs):
```
vmarkovtsev: Remove the default

sourced/ml/transformers/repo2weighted.py
```diff
@@ -1,16 +1,23 @@
...
         super().__init__(**kwargs)
         self.extractors = extractors
+        self.document_column = document_column
 
     def __call__(self, rows):
-        return rows.map(self.process_row)
+        res = rows \
```
vmarkovtsev: Same here, please `return` at once

sourced/ml/transformers/tfidf.py
```diff
@@ -0,0 +1,48 @@
...
+class TFIDF(Transformer):
+    """
+    Converts a single UAST into the weighted set (dictionary), where elements are strings
+    and the values are floats. The derived classes must implement uast_to_bag().
+    """
+    DEFAULT_DOCFREQ_THRESHOLD = 5
+
+    def __init__(self, tf, df, docfreq_threshold=None, **kwargs):
+        """
+        :param docfreq_threshold: The minimum number of occurrences of an element to be included \
```
vmarkovtsev: Would be nice to see what are `tf` and `df` too (including the type!)

sourced/ml/transformers/tfidf.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def _get_log_name(self):
+        return type(self).__name__
+
+    def __call__(self, head):
+        ndocs = self.df.count()
+        tfidf = self.df \
+            .map(lambda x: (x[0][1], (x[0][0], x[1]))) \
+            .join(self.tf) \
+            .map(lambda x: (x[1][0][0], x[0], logTF_log_IDF(x[1][1], x[1][0][1], ndocs)))
```
vmarkovtsev: At this time I cry blood. Is there any chance you make it accept and RDD of `Row`-s and use the names instead? Or `namedtuple`-s. Otherwise completely unmaintainable.

sourced/ml/algorithms/tf_idfs.py
```diff
@@ -0,0 +1,5 @@
+import numpy
+
+
+def logTF_log_IDF(tf, df, ndocs):
```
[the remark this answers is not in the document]
zurk: it is a function

sourced/ml/cmd_entries/repos2bow.py
zurk: calculate tfidf and get all values from rdd. what is wrong?

sourced/ml/algorithms/tf_idfs.py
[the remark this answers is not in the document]
vmarkovtsev: then make it lower case `log_tf_log_idf`

sourced/ml/cmd_entries/repos2bow.py
vmarkovtsev: Do not re-use `tfidf` for `tfidf.take`, very confusing

sourced/ml/transformers/repo2df.py
zurk: We you `processed ` in debug output (`processed.toDebugString().decode()`), so I can not join it with the upper line.

So, It will look like:
```python
    def __call__(self, rows):
        processed = rows.flatMap(self._process_row)
        if self.explained:
            self._log.info("toDebugString():\n%s", processed.toDebugString().decode())

        return processed \
            .flatMap(self._process_row) \
            .distinct() \
            .map(lambda x: (x[0], 1)) \
            .reduceByKey(operator.add)
```

Good?

vmarkovtsev: ok!

zurk:
@vmarkovtsev, It is ready to review and potential merge.

sourced/ml/cmd_entries/repos2bow.py
```diff
@@ -0,0 +1,47 @@
...
+        .link(UastDeserializer())
+    df_pipeline = uast_pipeline.link(Uast2DocFreq(extractors, document_column_name))
+    df = df_pipeline.execute()
+    tf_pipeline = uast_pipeline.link(Uast2TermFreq(extractors, document_column_name))
+    tf = tf_pipeline.execute()
+    tfidf = TFIDF(tf=tf, df=df).execute()
+
+    tokens, documents, values = zip(*tfidf.take(tfidf.count()))
+    log.info("Creating index mappings")
+    doc2id = {doc: doc_id for doc_id, doc in enumerate(sorted(list(set(documents))))}
```
vmarkovtsev: `sorted` does not need `list` inside.

Besides: why `set`? Is there a way they can contain duplicates?

sourced/ml/cmd_entries/repos2df.py
```diff
@@ -14,17 +14,17 @@ def repos2df_entry(args):
     extractors = [__extractors__[s](
         args.min_docfreq, **__extractors__[s].get_kwargs_fromcmdline(args))
         for s in args.feature]
+    document_column_name = "repository_id"
```
vmarkovtsev: Would be nice to declare a constant somewhere. E.g. new file `engine_constants.py` or just `engine.py` which already exists

sourced/ml/transformers/tfidf.py
```diff
@@ -0,0 +1,63 @@
...
+
+TFIDF_Row = namedtuple("TDIDF_Row", ["token", "document", "value"])
+
+
+class TFIDF(Transformer):
+    """
+    Converts a single UAST into the weighted set (dictionary), where elements are strings
+    and the values are floats. The derived classes must implement uast_to_bag().
+    """
+    DEFAULT_DOCFREQ_THRESHOLD = 5
```
vmarkovtsev: Would be nice to remove it here and take from BagsExtractor in `__init__`. Or vice versa

sourced/ml/cmd_entries/repos2bow.py
```diff
@@ -0,0 +1,47 @@
...
+from sourced.ml.models import BOW, OrderedDocumentFrequencies, DocumentFrequencies
+from sourced.ml.transformers import Engine, UastExtractor, UastDeserializer, Uast2DocFreq, \
+    Uast2TermFreq, HeadFiles, TFIDF
+from sourced.ml.utils import create_engine
+
+
+def repos2bow_entry(args):
+    log = logging.getLogger("repos2bow")
+    engine = create_engine("repos2bow-%s" % uuid4(), **args.__dict__)
+    document_column_name = "repository_id"
```
vmarkovtsev: Would be nice to declare a constant somewhere. E.g. new file engine_constants.py or just engine.py which already exists

sourced/ml/transformers/tfidf.py
```diff
@@ -0,0 +1,63 @@
+from collections import namedtuple
+import typing
+
+import pyspark
+
+from sourced.ml.algorithms import log_tf_log_idf
+from sourced.ml.transformers.transformer import Transformer
+
+TFIDF_Row = namedtuple("TDIDF_Row", ["token", "document", "value"])
```
vmarkovtsev: UpperCamelCase

sourced/ml/transformers/tfidf.py
```diff
@@ -0,0 +1,63 @@
...
+    def __call__(self, head):
+        ndocs = self.df.count()
+        docfreq_threhold = self.docfreq_threhold
+        df_pruned = self.df.filter(lambda x: x.value >= docfreq_threhold)
+        return self.tf \
+            .map(lambda x: (x.token, (x.document, x.value))) \
+            .join(df_pruned) \
+            .map(lambda x: TFIDF_Row(token=x[0],
+                                     document=x[1][0][0],
+                                     value=log_tf_log_idf(df=x[1][1], tf=x[1][0][1], ndocs=ndocs)))
```
vmarkovtsev: Still there are magical indices here :(

sourced/ml/transformers/uast2termfreq.py
```diff
@@ -0,0 +1,25 @@
...
+    def __init__(self, extractors, document_column, **kwargs):
+        super().__init__(**kwargs)
+        self.extractors = extractors
+        self.document_column = document_column
+
+    def __call__(self, rows):
+        return rows \
+            .flatMap(self.process_row) \
+            .reduceByKey(operator.add) \
+            .map(lambda x: TermFreqRow(document=x[0][0], token=x[0][1], value=x[1]))
```
vmarkovtsev: Would be great to use `Row` or `namedtuple` here too

sourced/ml/transformers/tfidf.py
zurk: `TFIDFRow` or `TfIdfRow` or `TFIDFRow`?

sourced/ml/cmd_entries/repos2bow.py
zurk: actually, I think to make it as a parameter for cmd entry point. We can choose `document`, `file`, etc. And it is a good idea to make a constant dictionary to map document type to column id.

sourced/ml/cmd_entries/repos2bow.py
zurk: yes, it contain duplicates, because each (tokens[i], documents[i], values[i]) is one matrix element.

sourced/ml/transformers/tfidf.py
vmarkovtsev: `TFIDFRow` - TFIDF is a common acronym

sourced/ml/cmd_entries/repos2bow.py
vmarkovtsev: Let's not overcomplicate the configuration, thank god this is not Java. A constant is enough

sourced/ml/transformers/tfidf.py
zurk: but I use it. What do you mean?

sourced/ml/transformers/uast2termfreq.py
zurk: but I use it. What do you mean?

sourced/ml/cmd_entries/repos2bow.py
vmarkovtsev: Is there a risk of explosion here given the size of the matrix? Then better to calc unique elements on Spark

sourced/ml/transformers/uast2termfreq.py
vmarkovtsev: I mean `x[0][0]` and friends

sourced/ml/cmd_entries/repos2bow.py
zurk: but do we want to use as document only repository?
We need it to use the same code for files as documents

vmarkovtsev: I don't see how this question is related to replacing hardcoded strings here with constants from a dedicated file

vmarkovtsev: OK, now I see

vmarkovtsev: For now - KISS. We will return to this question in the future

sourced/ml/cmd_entries/repos2bow.py
zurk: I think that not really, because anyway, we need all these values to save it to the disk, so anyway they should fit memory size.

vmarkovtsev: We need **indices** of tokens and documents, not their strings. And we need the lists of unique tokens nd documents. This will fit into memory. Individual, unindexed strings will explode and that is expected.

sourced/ml/transformers/uast2termfreq.py
zurk: oh, ok. about it and https://github.com/src-d/ml/pull/160#discussion_r161794558

I can not do here anything more since each operation will override my `namedtuple` (or `pyspark.Row`, what a little bit strange) with usual `tuple`, and I get `AttributeError: 'tuple' object has no attribute 'token'`. I do not find the solution to this problem.

One thing I can do is to use `.toDF()` at the end of `df` and `tf` calculation. In this case it will look better. One problem, we agreed that `Transformer` should return `rdd` or something iterable with tuples.

sourced/ml/transformers/tfidf.py
zurk: ok, I decide that it is better to use just `pyspark.Row` everywhere :)

sourced/ml/transformers/uast2termfreq.py
zurk: or not... I have one idea. will check.

zurk: Only like this:

```python
return self.tf \
            .map(lambda x: (x.token, Row(document=x.document, tf=x.value))) \
            .join(df_pruned) \
            .map(lambda x: Row(token=x[0],
                               document=x[1][0].document,
                               value=log_tf_log_idf(df=x[1][1], tf=x[1][0].tf, ndocs=ndocs)))
```

Not sure it is better...

vmarkovtsev: All right, if it is WONTFIX, then please comment every index

zurk:
@vmarkovtsev please review it and after I will squash everything and sign.

sourced/ml/cmd_entries/repos2bow.py
```diff
@@ -6,42 +6,46 @@
...
-from sourced.ml.utils import create_engine
+    Uast2TermFreq, HeadFiles, TFIDF, Cacher, Indexer
+from sourced.ml.utils import create_engine, EnineConstants
 
 
 def repos2bow_entry(args):
     log = logging.getLogger("repos2bow")
     engine = create_engine("repos2bow-%s" % uuid4(), **args.__dict__)
-    document_column_name = "repository_id"
+    document_column_name = EnineConstants.REPOSITORY_COLUMN_NAME
```
vmarkovtsev: **Engine** not Enine

vmarkovtsev: Also better REPOSITORY_ID_COLUMN_NAME - there are also URLs and other meta

vmarkovtsev: If you want to go wild you can make `EngineConstants.Columns.RepositoryId`

sourced/ml/transformers/indexer.py
```diff
@@ -0,0 +1,60 @@
...
+from pyspark.rdd import RDD
+from pyspark import Row
+
+from sourced.ml.transformers import Transformer
+
+
+class Indexer(Transformer):
+    """
+    Transformer that creates index mapping for one column in spark rdd.
+    It map all values in the column to its index.
```
vmarkovtsev: It maps

sourced/ml/transformers/indexer.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def __call__(self, rdd: RDD):
+        column_id = self.column
+        if isinstance(column_id, str):
+            column = rdd.map(lambda x: (getattr(x, column_id),))
+        else:
+            column = rdd.map(lambda x: (x[column_id],))
+
+        self._column_values = column \
+            .distinct() \
```
vmarkovtsev:
```
.distinct() \
.sortByKey() \
...
```

sourced/ml/transformers/indexer.py
```diff
@@ -0,0 +1,60 @@
...
+        :param column: column index or its name in pyspark.RDD for indexing.
+        :param kwargs:
+        """
+        super().__init__(**kwargs)
+        self.column = column
+        self._column2id = None
+        self._column_values = None
+
+    @property
+    def column2id(self):
```
vmarkovtsev: How about `__getitem__` instead?

sourced/ml/transformers/indexer.py
```diff
@@ -0,0 +1,60 @@
...
+    """
+    def __init__(self, column: typing.Union[int, str], **kwargs):
+        """
+        :param column: column index or its name in pyspark.RDD for indexing.
+        :param kwargs:
+        """
+        super().__init__(**kwargs)
+        self.column = column
+        self._column2id = None
+        self._column_values = None
```
vmarkovtsev: Naming:

```
self.column
self._column2id -> self._value_to_index
self._column_values -> self._values
``

File: sourced/ml/transformers/indexer.py [hunk 33]
```diff
@@ -0,0 +1,60 @@
...
+
+    @property
+    def column2id(self):
+        if self._column2id is None:
+            raise RuntimeError("column2id value not available. Run Indexer first.")
```
vmarkovtsev: AttributeError - it is kinda "not present"

sourced/ml/transformers/indexer.py
```diff
@@ -0,0 +1,60 @@
...
+        return self._column2id
+
+    @property
+    def column_values(self):
+        if self._column_values is None:
+            raise RuntimeError("column_values value not available. Run Indexer first.")
```
vmarkovtsev: AttributeError

sourced/ml/transformers/indexer.py
```diff
@@ -0,0 +1,60 @@
...
+
+        self._column_values = column \
+            .distinct() \
+            .map(lambda x: x[0]) \
+            .collect()
+
+        column2id = {d: i for i, d in enumerate(self._column_values)}
+        self._column2id = column2id
+
+        def index_column(row):
```
vmarkovtsev: This is suboptimal. PySpark will attempt to pickle this function. This is a scoped function which has `self` in the closure. This means `self` will be pickled too and subsequently the mappings. We only need `column2id` and do not need the second one.

Fix: `functools.partial` on a `@staticmethod` or a class.

sourced/ml/cmd_entries/repos2bow.py
```diff
@@ -6,42 +6,46 @@
...
     tf_pipeline = uast_pipeline.link(Uast2TermFreq(extractors, document_column_name))
     tf = tf_pipeline.execute()
-    tfidf = TFIDF(tf=tf, df=df).execute()
 
-    tokens, documents, values = zip(*tfidf.take(tfidf.count()))
-    log.info("Creating index mappings")
-    doc2id = {doc: doc_id for doc_id, doc in enumerate(sorted(list(set(documents))))}
-    token2id = {token: token_id for token_id, token in enumerate(sorted(list(set(tokens))))}
+    document_idexer = Indexer("document")
+    token_indexer = Indexer("token")
```
vmarkovtsev: Should be EngineConstants too

sourced/ml/transformers/indexer.py
zurk: Ok, it seems to be true, but if I use `self` inside this function. And in this case, pyspark fails to pickle it for some reason. So I do such things as `column_id = self.column` and so on. It seems that spark optimizes something and do not try to pickle self, if I don't use it inside. 

I did the following to check my words: add `__getstate__` to this class:
```python
    def __getstate__(*args, **kwargs):
        print("\nGEEEEEET STAAAAAATE\n")
```

Nothing was broke and I did not see the output. I will try to find some explanation in the documentation about such behavior. I prefer to leave it as is

sourced/ml/transformers/indexer.py
zurk: One thing, I am not sure it is the best idea. If someone decides to index not string values, or int values, `__getitem__` can have unexpected behavior.

I prefer to add `__getitem__` as a shortcut for `column2id` property. ok?

sourced/ml/cmd_entries/repos2bow.py
zurk: it is not from the engine. It is from `tdidf` Transformer. I create Row names on my own.

sourced/ml/transformers/indexer.py
vmarkovtsev: Such things are implementation specific. It may work, and may not in other interpreters, Spark versions, etc. Please be explicit. It's not hard to refactor.

sourced/ml/transformers/indexer.py
vmarkovtsev: `value_to_index`, yes

sourced/ml/cmd_entries/repos2bow.py
vmarkovtsev: OK, so place them in your class then

sourced/ml/transformers/indexer.py
zurk: it is ok to do as I did (http://spark.apache.org/docs/latest/rdd-programming-guide.html#passing-functions-to-spark). I Discuss it with @vmarkovtsev and decide to add comment.

PR #164, opened 2018-01-17
Add repos2ids command to collect identifiers

warenlg:


sourced/ml/__main__.py
```diff
@@ -66,6 +66,22 @@ def add_default_args(my_parser):
         choices=[ex.NAME for ex in extractors.__extractors__.values()],
         required=True, help="The feature extraction scheme to apply.")
 
+    repos2ids_parser = subparsers.add_parser(
+        "repos2ids", help="Convert source code to a token identifier stream."
```
vmarkovtsev: The output should be an ASDF model file with all the unique identifiers.

A stream is pipe or smth non-physical, it is not a file.

vmarkovtsev: Also please put a comment here how it is different from repos2df and why we need to invent it

sourced/ml/__main__.py
```diff
@@ -66,6 +66,22 @@ def add_default_args(my_parser):
...
+                           "The identifiers are splittable on special characters"
+                           "or case changes")
+    repos2ids_parser.set_defaults(handler=repos2ids_entry)
+    add_engine_args(repos2ids_parser)
+    repos2ids_parser.add_argument(
+        "-r", "--repositories", required=True,
+        help="The path to the repositories.")
+    repos2ids_parser.add_argument(
+        "--token-stream", required=True,
```
vmarkovtsev: -o, --output

sourced/ml/__main__.py
```diff
@@ -66,6 +66,22 @@ def add_default_args(my_parser):
...
+        help="[OUT] The path to the token stream of identifiers.")
+    repos2ids_parser.add_argument(
+        "--vocabulary-size", default=10000000, type=int,
+        help="The maximum vocabulary size.")
```
vmarkovtsev: You also need to filter only those which are more than 1 part. We should add a flag.

warenlg: @vmarkovtsev , I don't understand the purpose of such a flag. I thought there was no other option: we are always going to filter tokens with more than 1 part.

vmarkovtsev: The name of your command is `repos2ids`. You generate unique identifiers from source code. You do not dump them splitted or X-Y-ed. Filtering 1-part tokens is a part of your specific application to train the model. We can perform that action later and can perform it here to save some time and space. However, we do not have to.

warenlg: Yes I get it now, thx!

zurk:
@vmarkovtsev, @warenlg Can we use another name? `repos2ids` is too confusing for me and do not explain a lot. 
Something like `repos2id_stream` or `repos_id2stream`

What do you think guys?

vmarkovtsev:
There is no "stream". A stream is this:
stream

The command indeed extracts identifiers from repositories. In contrast to docfreq, it does not try to calc any frequencies and does not write any model - it should dump to stdout. Also it can work without Babelfish.

vmarkovtsev:
(We are open to other suggestions still)

zurk:
http://en.cppreference.com/w/cpp/header/iostream 
🤔
Stream is a good word.
If you do not like it lets use something like `repos_id2stdout`

vmarkovtsev:
@zurk Don't mix the way it works and what it is supposed to do. The goal is to extract the list of unique identifiers. It can write it to a DB, to a stream, to a file or to whatever else.

vmarkovtsev:
@warenlg You definitely need to rebase here

vmarkovtsev:
@warenlg I plan to read this tomorrow. For now - there is a conflict which needs to be fixed.

warenlg:
Conflicts resolved

sourced/ml/__main__.py
```diff
@@ -57,6 +57,28 @@ def add_parser(name, help):
...
+    repos2ids_parser = subparsers.add_parser(
+        "repos2ids", help="Convert source code to a bag of identifiers.")
+    repos2ids_parser.set_defaults(handler=repos2ids_entry)
+    add_engine_args(repos2ids_parser)
+    repos2ids_parser.add_argument(
+        "-r", "--repositories", required=True,
+        help="The path to the repositories.")
+    repos2ids_parser.add_argument(
+        "-o", "--output", required=True,
+        help="[OUT] The path to the output csv file of identifiers, compressed in gz format.")
```
vmarkovtsev: output **CSV** file **with** identifiers

You should not compress it unless the file name ends with `.gz`

sourced/ml/__main__.py
```diff
@@ -57,6 +57,28 @@ def add_parser(name, help):
...
+    repos2ids_parser.add_argument(
+        "--split", action="store_true",
+        help="Flag that specifies if you want to filter specific identifiers."
```
vmarkovtsev: Make it shorter:
Be shorter:
```
Enables filtering identifiers...
```

Honestly, I don't get the help message at all. What do you mean?

sourced/ml/__main__.py
```diff
@@ -57,6 +57,28 @@ def add_parser(name, help):
...
+             "Default value makes you filter identifiers which are easy splittable:"
+             "that is to say on special characters or case changes")
+    repos2ids_parser.add_argument(
+        "--idfreq", action="store_true",
+        help="Flaf that adds identifier frequencies to the output CSV file."
```
vmarkovtsev:
```
Adds identifier frequencies to the output CSV file.
```

sourced/ml/algorithms/token_parser.py
```diff
@@ -9,13 +9,16 @@ class TokenParser:
...
+    MIN_SPLIT_LENGTH = 3  #: We do not split source code identifiers shorter than this value.
 
-    def __init__(self, stem_threshold=STEM_THRESHOLD, max_token_length=MAX_TOKEN_LENGTH):
+    def __init__(self, stem_threshold=STEM_THRESHOLD, max_token_length=MAX_TOKEN_LENGTH,
+                 min_split_length=MIN_SPLIT_LENGTH):
         self._stemmer = Stemmer.Stemmer("english")
         self._stemmer.maxCacheSize = 0
         self._stem_threshold = stem_threshold
         self._max_token_length = max_token_length
+        self.min_split_length = min_split_length
```
vmarkovtsev: Since the rest of the values are protected, the idea was to add properties later once we have time. You should make this protected too, and add getters/setters with type and range checks if you have time.

sourced/ml/cmd_entries/repos2ids.py
```diff
@@ -0,0 +1,24 @@
...
+
+from sourced.ml.transformers import Ignition, ContentExtractor, ContentDeserializer, \
+    Content2Ids, HeadFiles, Cacher
+from sourced.ml.utils import create_engine, EngineConstants
+
+
+def repos2ids_entry(args):
+    log = logging.getLogger("repos2ids")
+    engine = create_engine("repos2ids-%s" % uuid4(), **args.__dict__)
+    documents_column_name = [EngineConstants.Columns.RepositoryId,
```
vmarkovtsev: `documents_column_name` -> `column_names`

sourced/ml/cmd_entries/repos2ids.py
```diff
@@ -0,0 +1,24 @@
...
+                             EngineConstants.Columns.PathId]
+    ids_transformer = Content2Ids(args, documents_column_name)
+
+    pipeline = Ignition(engine) \
```
vmarkovtsev:
```
ids = Ignition(engine) \
    .link(HeadFiles()) \
    # ...
    .execute()
```

sourced/ml/cmd_entries/repos2ids.py
```diff
@@ -0,0 +1,24 @@
...
+        .link(HeadFiles()) \
+        .link(ContentExtractor()) \
+        .link(ContentDeserializer()) \
+        .link(ids_transformer)
+    ids = pipeline.execute()
+
+    log.info("Writing %s", args.output)
+    ids_transformer.save(ids)
```
vmarkovtsev: I would make it a static method and then `.link(Content2Ids(args, column_names))`

sourced/ml/tests/tfidf_data.py
```diff
@@ -14,6 +17,143 @@ def readonly(_dict: dict):
...
+             "lang": ["Python", "Scala", "Go", "Java", "C++", "JavaScript"], },
+            {"uast": [0, 1, 2, 3, 4, 5],
+             "content": ["method_base", "QueryBuilder", "resolvedFiler",
+                         "Flag", "MinhashCuda", "baseConf"],
+             "file": [0, 1, 2, 3, 4, 5],
+             "document": [0, 0, 0, 1, 1, 1],
+             "lang": ["Python", "Scala", "Go", "Java", "C++", "JavaScript"], },
+            ]
+
+ids_result = [{Row(token='method_base', token_split='method base'),
```
vmarkovtsev: Please replace single quotes with double

sourced/ml/transformers/basic.py
```diff
@@ -113,6 +113,18 @@ def __call__(self, files):
         return uasts
 
 
+class ContentExtractor(Transformer):
```
vmarkovtsev: This is not basic, you should move it to a separate file or to `content2ids.py`

sourced/ml/transformers/basic.py
```diff
@@ -113,6 +113,18 @@ def __call__(self, files):
...
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+
+    def __call__(self, files):
+        files = files.dropDuplicates(("blob_id",)).filter("is_binary = 'false'")
+        classified = files.classify_languages().filter("lang is not null")
```
vmarkovtsev: Chain them

```
return files \
    .dropDuplicates(("blob_id",)) \
    .filter("is_binary = 'false'") \
    .classify_languages() \
    .filter("lang is not null") \
    .where(functions.length(functions.col("content")) > 0)
```

sourced/ml/transformers/basic.py
```diff
@@ -160,3 +172,17 @@ def deserialize_uast(self, row):
         row_dict = row.asDict()
         row_dict["uast"] = self.parse_uast(row.uast[0])
         yield Row(**row_dict)
+
+
+class ContentDeserializer(Transformer):
```
vmarkovtsev: You don't need this at all. The only reason we have UastDeserializer is that UASTs cannot be pickled - it is happening while persisting the pipeline, so we have to cache bytes and deserialize them every time.

In your case it is already bytes and they are perfectly pickleable.

sourced/ml/transformers/content2ids.py
```diff
@@ -0,0 +1,146 @@
...
+        filenames = []
+
+        def __init__(self, **options):
+            super(Content2Ids.FormatterProxy, self).__init__(**options)
+            self.callback = options["callback"]
+
+        def format(self, tokensource, outfile):
+            self.callback(tokensource)
+
+    def __init__(self, args, documents_column, **kwargs):
```
vmarkovtsev: Transformers should never accept bulk `argparse.Namespace`-s. You need to define the set of args/kwargs and specify them explicitly.

Also would be nice to have type annotations.

sourced/ml/transformers/content2ids.py
```diff
@@ -0,0 +1,146 @@
...
+        processed = rows.flatMap(self._process_row).persist()
+
+        if self.idfreq:
+            num_repos_processed = processed \
+                .map(lambda x: (x[0], x[1][0])) \
+                .distinct()
+            num_repos_reduced = self.reduce_rows(num_repos_processed)
+            list_RDDs.append(num_repos_reduced)
+
+            num_files_processed = processed \
```
vmarkovtsev: I see duplicated code

sourced/ml/transformers/content2ids.py
```diff
@@ -0,0 +1,146 @@
...
+                .distinct()
+            num_files_reduced = self.reduce_rows(num_files_processed)
+            list_RDDs.append(num_files_reduced)
+
+            num_occ_reduced = self.reduce_rows(processed)
+            list_RDDs.append(num_occ_reduced)
+
+            return processed \
+                .map(lambda x: x[0]) \
+                .context.union(list_RDDs) \
```
vmarkovtsev: Why do you need `map` before `context.union`?

sourced/ml/transformers/content2ids.py
```diff
@@ -0,0 +1,146 @@
...
+                    token_split=" ".join(TokenParser(min_split_length=1).split(x[0])),
+                    num_repos=x[1][0],
+                    num_files=x[1][1],
+                    num_occ=x[1][2]))
+        else:
+            return processed \
+                .map(lambda x: x[0]) \
+                .distinct() \
+                .map(lambda x: Row(
+                    token=x,
```
vmarkovtsev: pep8 indentation

sourced/ml/transformers/content2ids.py
```diff
@@ -0,0 +1,146 @@
...
+                    token_split=" ".join(TokenParser(min_split_length=1).split(x))))
+
+    def reduce_rows(self, rows):
+        return rows \
+            .map(lambda x: (x[0], 1)) \
+            .reduceByKey(operator.add)
+
+    def _process_row(self, row):
+        self.names = []
+        repo_id = getattr(row, self.documents_column[0])
```
vmarkovtsev: This is why `documents_columns` must be a `namedtuple`

sourced/ml/transformers/content2ids.py
```diff
@@ -0,0 +1,146 @@
...
+        file_id = getattr(row, self.documents_column[1])
+        path = os.path.join(repo_id, file_id)
+        code = row.content
+        try:
+            lexer = get_lexer_by_name(self.linguist2pygments[row.lang][0])
+            pygments.highlight(code, lexer, self.FormatterProxy(callback=self.process_tokens))
+        except ClassNotFound:
+            lexer = get_lexer_by_name(self.linguist2pygments[row.lang][1])
+            pygments.highlight(code, lexer, self.FormatterProxy(callback=self.process_tokens))
```
vmarkovtsev: Code is duplicated

```
for i in (0, 1):
    try:
        # ...
        break
    except ClassNotFound:
        continue
```

sourced/ml/transformers/content2ids.py
```diff
@@ -0,0 +1,146 @@
...
+
+    def process_tokens(self, tokens):
+        """
+        Filter tokens of type "Name" and which are splittable
+        according to :class: 'TokenParser' rules
+        """
+        for _type, token in tokens:
+            if _type[0] == "Name":
+                if self.split:
+                    if len(list(TokenParser(min_split_length=1).split(token))) > 1:
```
vmarkovtsev: `len` + `list` is a bad way to calc length of iterators. `sum(1 for _ in ...)` is much better because it does not materialize all the items

sourced/ml/transformers/content2ids.py
```diff
@@ -0,0 +1,146 @@
...
+        self.output = args.output
+
+    def __call__(self, rows):
+        list_RDDs = []
+        self.build_mapping()
+        processed = rows.flatMap(self._process_row).persist()
+
+        if self.idfreq:
+            num_repos_processed = processed \
+                .map(lambda x: (x[0], x[1][0])) \
```
vmarkovtsev: Would be nice to have a comment which explains the initial structure of `x`

sourced/ml/transformers/content2ids.py
```diff
@@ -0,0 +1,146 @@
...
+        self.documents_column = documents_column
+        self.linguist2pygments = {}
+        self.split = args.split
+        self.idfreq = args.idfreq
+        self.output = args.output
+
+    def __call__(self, rows):
+        list_RDDs = []
+        self.build_mapping()
+        processed = rows.flatMap(self._process_row).persist()
```
vmarkovtsev: `persist()` may not be executed inside a Transformer unless it is a Cacher. Split the class into two and insert Cacher in the middle. Please note: in your case there should be only one Cacher for the whole pipeline tree.

sourced/ml/transformers/content2ids.py
```diff
@@ -0,0 +1,146 @@
...
+                    if len(list(TokenParser(min_split_length=1).split(token))) > 1:
+                        self.names.append(token)
+                else:
+                    self.names.append(token)
+
+    def build_mapping(self):
+        """
+        Builds the mapping between linguist languages and pygments names for lexers.
+        """
+        with open("doc/languages.yml") as f:
```
vmarkovtsev: `doc` is for the documentation. The file must be placed somewhere else, e.g. in `path.dirname(__file__)`

Do not forget to include it into the package in `setup.py` (and check that it is included!).

sourced/ml/transformers/content2ids.py
```diff
@@ -0,0 +1,146 @@
...
+        """
+        for _type, token in tokens:
+            if _type[0] == "Name":
+                if self.split:
+                    if len(list(TokenParser(min_split_length=1).split(token))) > 1:
+                        self.names.append(token)
+                else:
+                    self.names.append(token)
+
+    def build_mapping(self):
```
vmarkovtsev: Can be made static and the mappings too

sourced/ml/transformers/content2ids.py
```diff
@@ -0,0 +1,146 @@
...
+            pygments_langs |= set(lang_declensions)
+
+        for lang in linguist_langs:
+            lang_names = linguist_langs.get(lang, (set(),))[0]
+            inter = list(lang_names.intersection(pygments_langs))
+            if inter:
+                self.linguist2pygments[lang] = inter
+
+    def save(self, id_rdd):
+        with gzip.open(self.output, "w") as g:
```
vmarkovtsev: Remember my comment in `__main__.py` about gzip

sourced/ml/transformers/content2ids.py
```diff
@@ -0,0 +1,146 @@
...
+            columns_names = ["token", "token_split"]
```
vmarkovtsev: I think it is a nice opportunity to use https://pandas.pydata.org/pandas-docs/stable/generated/pandas.DataFrame.to_csv.html and erase this tricky ad-hoc code

warenlg:
@vmarkovtsev  I'd say it looks better now

setup.py
```diff
@@ -31,13 +31,15 @@
...
+                      "pygments>=2.2.0,<3.0",
                       "humanize>=0.5.0",
                       "parquet>=1.2,<2.0"] + typing,
     extras_require={
         "tf": ["tensorflow>=1.0,<2.0"],
         "tf_gpu": ["tensorflow-gpu>=1.0,<2.0"],
     },
-    package_data={"": ["LICENSE", "README.md"]},
+    package_data={"": ["LICENSE", "README.md"],
+                  "sourced": ["ml/transformers/languages.yml"]},
```
vmarkovtsev: Is there a test which checks that this file is present?

sourced/ml/cmd_entries/repos2ids.py
```diff
@@ -1,24 +1,25 @@
...
 from sourced.ml.utils import create_engine, EngineConstants
 
 
 def repos2ids_entry(args):
     log = logging.getLogger("repos2ids")
     engine = create_engine("repos2ids-%s" % uuid4(), **args.__dict__)
-    documents_column_name = [EngineConstants.Columns.RepositoryId,
-                             EngineConstants.Columns.PathId]
-    ids_transformer = Content2Ids(args, documents_column_name)
+    Column = NamedTuple("Column", [("repo_id", str), ("file_id", str)])
```
vmarkovtsev: Column -> Columns

This should be defined and required inside Content2Ids

sourced/ml/cmd_entries/repos2ids.py
```diff
@@ -1,24 +1,25 @@
...
+    language_mapping = Content2Ids.build_mapping()
```
vmarkovtsev: This should be a static and lazily created variables in Content2Ids - that is, initialized at the first request.

sourced/ml/transformers/content2ids.py
```diff
@@ -26,94 +28,84 @@ def __init__(self, **options):
         def format(self, tokensource, outfile):
             self.callback(tokensource)
 
-    def __init__(self, args, documents_column, **kwargs):
+    def __init__(self, language_mapping: Dict, column_names: NamedTuple,
```
vmarkovtsev: put the exact type of `column_names` here

sourced/ml/transformers/content2ids.py
```diff
@@ -26,94 +28,84 @@ def __init__(self, **options):
...
-                .distinct()
-            num_files_reduced = self.reduce_rows(num_files_processed)
-            list_RDDs.append(num_files_reduced)
-
-            num_occ_reduced = self.reduce_rows(processed)
-            list_RDDs.append(num_occ_reduced)
-
+            for i in (0, 1):
+                # initial structure of x: (identifier, (repositoryId, filepath))
+                freq_processed = processed \
```
vmarkovtsev: I bet you can fit it to the single line

sourced/ml/transformers/content2ids.py
```diff
@@ -26,94 +28,84 @@ def __init__(self, **options):
...
+        self.split = split
+        self.idfreq = idfreq
 
-    def __call__(self, rows):
+    def __call__(self, rows: RDD):
         list_RDDs = []
-        self.build_mapping()
-        processed = rows.flatMap(self._process_row).persist()
-
+        processed = rows.flatMap(self._process_row)
```
vmarkovtsev: And this is the problem because you will calculate `self_process_row` every time. That's why I suggested to split this class and put a Cacher in the middle.

sourced/ml/algorithms/token_parser.py
vmarkovtsev: This is still true

sourced/ml/cmd_entries/repos2ids.py
```diff
@@ -0,0 +1,23 @@
...
+    log = logging.getLogger("repos2ids")
+    engine = create_engine("repos2ids-%s" % uuid4(), **args.__dict__)
+
+    ids = Ignition(engine) \
+        .link(HeadFiles()) \
+        .link(ContentExtractor()) \
+        .link(ContentProcess(args.split)) \
+        .link(Cacher.maybe(args.persist)) \
+        .link(Content2Ids(args.idfreq)) \
+        .execute()
```
vmarkovtsev: This is sexy now

sourced/ml/transformers/content2ids.py
```diff
@@ -0,0 +1,148 @@
...
+class Content2Ids(Transformer):
+
+    def __init__(self, idfreq: bool, **kwargs):
+        super().__init__(**kwargs)
+        self.idfreq = idfreq
+
+    def __call__(self, rows: RDD):
+        list_RDDs = []
+        if self.idfreq:
+            for i in (0, 1):
```
vmarkovtsev: Please make either branch of this giant `if` a separate function in the next PR

sourced/ml/transformers/content2ids.py
```diff
@@ -0,0 +1,148 @@
...
+
+        for lang in linguist_langs:
+            lang_names = linguist_langs.get(lang, (set(),))[0]
+            inter = list(lang_names.intersection(pygments_langs))
+            if inter:
+                linguist2pygments[lang] = inter
+        return linguist2pygments
+
+
+class Content2Ids(Transformer):
```
vmarkovtsev: Let's rename it to `IdentifiersToDataset` in the next PR

sourced/ml/transformers/content2ids.py
```diff
@@ -0,0 +1,148 @@
...
+from pygments.lexers import get_lexer_by_name, ClassNotFound
+from pyspark import RDD, Row
+from pyspark.sql import functions
+
+from sourced.ml.algorithms import TokenParser
+from sourced.ml.transformers import Transformer
+from sourced.ml.utils import EngineConstants
+
+
+class ContentProcess(Transformer):
```
vmarkovtsev: Let's rename it to `ContentToIdentifiers` in the next PR

vmarkovtsev:
Good job! The rest is not critical and can be addressed later.

PR #168, opened 2018-01-23
Refactor repo2coocc entry point

zurk:
- Add min_docfreq parameter
- Add saving for document frequency model
- Add Extractors usage
- Delete outdated token_mapper transformer. We have Indexer now

Signed-off-by: Konstantin Slavnov konstantin@sourced.tech

PR #169, opened 2018-01-24
Fix id2vec_preproc to work with new repo2coocc

zurk:
PTAL https://github.com/src-d/ml/pull/168 first.


sourced/ml/__main__.py
```diff
@@ -139,8 +138,8 @@ def add_default_args(my_parser):
         help="Combine row and column embeddings together and write them to an .asdf.",
         formatter_class=ArgumentDefaultsHelpFormatterNoNone)
     id2vec_postproc_parser.set_defaults(handler=postprocess_id2vec)
-    id2vec_postproc_parser.add_argument("swivel_output_directory")
-    id2vec_postproc_parser.add_argument("result")
+    id2vec_postproc_parser.add_argument("--swivel-output-directory")
```
vmarkovtsev: It was without `--` so that the arguments were free. Now we have to type `--swivel-output-directory` every time. Also they are not required now. Add `-i`, `-o` shorthands, mark as required and add help.

sourced/ml/cmd_entries/preprocess_id2vec.py
```diff
@@ -22,25 +30,9 @@ def preprocess_id2vec(args):
...
-        try:
-            model = Cooccurrences().load(source=path)
-        except ValueError:
-            skipped += 1
-            log.warning("Skipped %s", path)
-            continue
-        for w in model.tokens:
-            all_words[w] += 1
+    df_model = DocumentFrequencies().load(source=args.docfreq)
+    # TODO: Add dependency check if it exists.
```
vmarkovtsev: Go ahead and add it now

sourced/ml/__main__.py
zurk: ok, I change it just to have the same parameter style in all sourced ml

PR #172, opened 2018-01-29
Add Sampler Transformer

zurk:
It is needed for debug purposes (speed up code execution) mostly.


sourced/ml/transformers/basic.py
```diff
@@ -5,6 +5,17 @@
 from sourced.ml.transformers.transformer import Transformer
 
 
+class Sampler(Transformer):
```
vmarkovtsev: Please add a docstring which says what this does

PR #178, opened 2018-02-22
Update docker file

zurk:
Now it is based on our spark docker file and can be used as spark worker.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:16.04
+FROM srcd/spark:2.2.0_v2
 
 ADD requirements.txt setup.py package/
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -29,5 +29,7 @@ RUN pip3 install --no-cache-dir ./package && rm -rf package
 EXPOSE 8000
 ENV BROWSER /browser
 ENV LC_ALL en_US.UTF-8
+ENV PYSPARK_PYTHON python3
+ENV PYSPARK_PYTHON_DRIVER python3
```
vmarkovtsev: This should be inside srcd/spark:2.2.0_v2 no?

PR #180, opened 2018-02-22
 Improvements: fixes, better feature extraction

EgorBu:
Modify features extraction algorithms
Fix ParquetLoader
Modify separators a bit

### DIFF
--- a/sourced/ml/algorithms/uast_struct_to_bag.py
+++ b/sourced/ml/algorithms/uast_struct_to_bag.py
@@ -5,6 +5,8 @@ from sourced.ml.algorithms.uast_ids_to_bag import FakeVocabulary, Uast2BagBase
 
 
 class Uast2StructBagBase(Uast2BagBase):
+    SEP = ">"
+
     def __init__(self, stride, seq_len, node2index=None):
         self._node2index = node2index if node2index is not None else FakeVocabulary()
         self._stride = stride
@@ -26,10 +28,10 @@ class UastSeq2Bag(Uast2StructBagBase):
     """
     DFS traversal + preserves the order of node children.
     """
-    def __init__(self, stride=1, seq_len=5, node2index=None, sep="|"):
+
+    def __init__(self, stride=1, seq_len=5, node2index=None):
         _node2index = Node2InternalType() if node2index is None else node2index
         super().__init__(stride, seq_len, _node2index)
-        self.sep = sep
 
     def _uast2sequence(self, root):
         sequence = []
@@ -54,7 +56,7 @@ class UastSeq2Bag(Uast2StructBagBase):
 
         for seq_len in self._seq_lens:
             for i in range(0, len(node_sequence) - seq_len + 1, self._stride):
-                key = self.sep.join(node_sequence[i:i + seq_len])
+                key = self.SEP.join(node_sequence[i:i + seq_len])
                 bag[key] += 1
         return bag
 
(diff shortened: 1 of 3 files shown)


sourced/ml/transformers/basic.py
```diff
@@ -123,6 +124,7 @@ def __init__(self, fields: Union[list, tuple], **kwargs):
     def __call__(self, df: DataFrame) -> DataFrame:
         res = df.select(self.fields)
         if self.explained:
+            self._log.info("toDebugString():\n%s", res.explain(True).decode())
```
vmarkovtsev: I think `res.explain()` prints to stdout, also is very impresise and loses the previous pipeline. Should be avoided.

sourced/ml/transformers/basic.py
```diff
@@ -134,17 +136,28 @@ def __init__(self, save_loc, **kwargs):
 
     def __call__(self, df: DataFrame):
         if self.explained:
+            self._log.info("toDebugString():\n%s", df.explain(True).decode())
```
vmarkovtsev: same here

sourced/ml/transformers/basic.py
```diff
@@ -134,17 +136,28 @@ def __init__(self, save_loc, **kwargs):
...
 
-    def __call__(self, df):
-        return self.session.read.parquet(self.save_loc)
+    def __getstate__(self):
+        state = super().__getstate__()
+        del state["session"]
+        return state
+
+    def __call__(self, _):
+        if issubclass(type(self.paths), list):
```
vmarkovtsev: why not `isinstance(self.paths, (list, tuple))`?

sourced/ml/transformers/basic.py
```diff
@@ -134,17 +136,28 @@ def __init__(self, save_loc, **kwargs):
...
+            return self.session.read.parquet(*self.paths)
+        elif issubclass(type(self.paths), str):
```
vmarkovtsev: You already returned, no `elif` is needed

sourced/ml/transformers/basic.py
```diff
@@ -153,13 +166,19 @@ def __setstate__(self, state):
         from bblfsh import Node
         self.parse_uast = Node.FromString
 
-    def __call__(self, rows: RDD) -> RDD:
+    def __call__(self, rows: Union[RDD, DataFrame]) -> RDD:
+        if isinstance(rows, DataFrame):
```
vmarkovtsev: Deserializer should always require an RDD. If it is a dataframe in your case, change the previous transfomer to return an rdd.

sourced/ml/transformers/basic.py
```diff
@@ -153,13 +166,19 @@ def __setstate__(self, state):
...
     def deserialize_uast(self, row: Row):
         if not row[EngineConstants.Columns.Uast]:
             return
         row_dict = row.asDict()
-        row_dict[EngineConstants.Columns.Uast] = self.parse_uast(
-            row[EngineConstants.Columns.Uast][0])
+        try:
+            row_dict[EngineConstants.Columns.Uast] = self.parse_uast(
+                row[EngineConstants.Columns.Uast][0])
+        except Exception as e:
```
vmarkovtsev: The exception during parsing a UAST is impossible. If it happens, we need to fix the problem itself.

sourced/ml/transformers/uast2bag_features.py
```diff
@@ -9,14 +9,18 @@
 
 
 class UastRow2Document(Transformer):
+    sep1 = "//"
```
vmarkovtsev: Name these better and capitalize

sourced/ml/transformers/basic.py
EgorBu: according to http://spark.apache.org/docs/2.2.0/api/python/pyspark.sql.html#pyspark.sql.DataFrameReader.parquet it can be either list of str (it's a bit confusing in their documentation)

sourced/ml/algorithms/uast_struct_to_bag.py
```diff
@@ -5,6 +5,8 @@
 
 
 class Uast2StructBagBase(Uast2BagBase):
+    SEP = "|"
```
vmarkovtsev: Do you think ">" is a better separator? We kind of point to the next node.

sourced/ml/transformers/basic.py
```diff
@@ -139,12 +139,22 @@ def __call__(self, df: DataFrame):
...
 
-    def __call__(self, df):
-        return self.session.read.parquet(self.save_loc)
+    def __getstate__(self):
+        state = super().__getstate__()
+        del state["session"]
+        return state
+
+    def __call__(self, _):
+        if issubclass(type(self.paths), (list, tuple)):
```
vmarkovtsev: `isinstance` please

sourced/ml/algorithms/uast_struct_to_bag.py
EgorBu: makes sense

PR #181, opened 2018-02-23
Wrap import of tensorflow to avoid import error

EgorBu:
Avoid import error on systems without `tensorflow`.

### DIFF
--- a/sourced/ml/cmd_entries/preprocess_id2vec.py
+++ b/sourced/ml/cmd_entries/preprocess_id2vec.py
@@ -3,7 +3,10 @@ import os
 import shutil
 
 import numpy
-import tensorflow as tf
+try:
+    import tensorflow as tf
+except ImportError as e:
+    logging.warn("Tensorflow is not installed, dependent functionality is unavailable.")
 
 from modelforge.progress_bar import progress_bar
 from sourced.ml.algorithms.id_embedding import _extract_coocc_matrix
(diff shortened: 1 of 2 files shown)


sourced/ml/algorithms/swivel.py
```diff
@@ -59,47 +59,52 @@
...
-flags.DEFINE_float("num_epochs", 40, "Number epochs to train for")
-flags.DEFINE_float("per_process_gpu_memory_fraction", 0,
-                   "Fraction of GPU memory to use, 0 means allow_growth")
-flags.DEFINE_integer("num_gpus", 0,
-                     "Number of GPUs to use, 0 means all available")
-flags.DEFINE_string("logs", "",
-                    "Path for TensorBoard logs (empty value disables them)")
-
-FLAGS = flags.FLAGS
+try:
```
vmarkovtsev: We cannot do this because we take this file from https://github.com/src-d/tensorflow-swivel

Wrap it on the upper level

sourced/ml/cmd_entries/preprocess_id2vec.py
```diff
@@ -3,7 +3,10 @@
 import shutil
 
 import numpy
-import tensorflow as tf
+try:
+    import tensorflow as tf
+except ImportError as e:
+    logging.warn("Tensorflow is not installed, related functionality is unavailable.")
```
vmarkovtsev: related -> dependent

vmarkovtsev:
@EgorBu pep8

PR #182, opened 2018-02-26
Add "repo" extraction mode

vmarkovtsev:


vmarkovtsev:
@zurk Please have a look, it may break some of the stuff you need.

The change is in the way we pump uasts. Before: `row.uast[0]` After: `row.uast` and then iterate over uasts in the list. In other words, we allow having several uasts in a document. I had to do that to support repository-level bags.

PR #186, opened 2018-02-27
Fix pyspark requirements

zurk:
We have pyspark requirements from `sourced-engine` package: 
https://github.com/src-d/engine/blob/master/python/setup.py#L29

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,5 +6,5 @@ sourced-engine>=0.5.1,<0.6
 parquet>=1.2,<2.0
 numpy>=1.14
 humanize>=0.5.0,<0.6
-pyspark==2.2.0.post0
+pyspark==2.2.0
 pygments>=2.2.0,<3.0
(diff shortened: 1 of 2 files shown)


vmarkovtsev:
Not so easy. Our cluster env uses 2.2.0 everywhere.

zurk:
mmm. ok 
I try to find another solution.
Just FYI.
To install pyspark you need to specify version as is: `pyspark==2.2.0.post0`
but it installs as just `2.2.0` and after says that `pkg_resources.DistributionNotFound: The 'pyspark==2.2.0.post0' distribution was not found and is required by sourced-ml`... :/

zurk:
now it works fine.

vmarkovtsev:
@zurk No it's not

```
  Could not find a version that satisfies the requirement pyspark==2.2.0 (from -r requirements.txt (line 9)) (from versions: 2.1.2, 2.2.0.post0, 2.2.1)
No matching distribution found for pyspark==2.2.0 (from -r requirements.txt (line 9))
```

I guess we can close this.

PR #189, opened 2018-03-01
Fix pyspark version

zurk:
Updated PR for https://github.com/src-d/ml/pull/186#event-1498671266
I really do not like this messages
```
pkg_resources.DistributionNotFound: The 'pyspark==2.2.0.post0' distribution was not found and is required by sourced-ml
```
:)

Now it should work

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,5 +6,5 @@ sourced-engine>=0.5.1,<0.6
 parquet>=1.2,<2.0
 numpy>=1.14
 humanize>=0.5.0,<0.6
-pyspark==2.2.0.post0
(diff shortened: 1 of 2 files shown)


vmarkovtsev:
@zurk Where and under which circumstances do you see this error message?

zurk:
when I try to run @bzz example from here: https://github.com/src-d/apollo/issues/18#issue-291925951 that the reason why I mentioned this PR on the issue.

vmarkovtsev:
@zurk The conflict you have posted is your personal local conflict - it says that you already have 2.2.0 installed and it does not match.

PR #193, opened 2018-03-13
Add deployment via travis

zurk:
The same way as in https://github.com/src-d/minhashcuda

there are 2 stages: testing, using python 3.4,5,6 and deploying. 
`python3 setup.py bdist_wheel` creates `sourced_ml-0.4.4-py3-none-any.whl` file, so there is no need to use `auditwheel` tool. 

Also I change `docker exec -it bblfshd bblfshctl driver install --all` to `docker exec -it bblfshd bblfshctl driver install python bblfsh/python-driver` because we need only one driver for testing and the error `API rate limit exceeded` inside Travis.
```
[0K$ docker exec -it bblfshd bblfshctl driver install --all
Error, GET https://api.github.com/orgs/bblfsh/repos?page=1&per_page=100&type=public: 403 API rate limit exceeded for 35.184.96.71. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.) [rate reset in 27m15s]
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ addons:
 stages:
   - test
   - name: deploy
+    # require any tag name to deploy
     if: tag =~ .*
 _install: &_install
   - gimme 1.8
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,10 @@ addons:
       - libboost-all-dev
       - libxml2-dev
       - libsnappy-dev
+stages:
+  - test
+  - name: deploy
+    if: tag =~ .*
 _install: &_install
   - gimme 1.8
   - source ~/.gimme/envs/latest.env
@@ -26,6 +30,12 @@ _install: &_install
   - pip install -e .
 _coverage: &_coverage
   - SCRIPT="coverage run --concurrency=multiprocessing -m unittest discover && coverage combine"
+_deploy: &_deploy
+  provider: script
+  script: twine upload dist/*py3-none-any* -u $PYPI_LOGIN -p $PYPI_PASS
+  skip_cleanup: true
+  on:
+    tags: true
 matrix:
   include:
     - python: 3.4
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -12,6 +12,10 @@ addons:
       - libboost-all-dev
       - libxml2-dev
       - libsnappy-dev
+stages:
+  - test
+  - name: deploy
+    if: tag =~ .*
```
vmarkovtsev: Does yaml support comments? Would be nice to explain this monster

.travis.yml
```diff
@@ -42,11 +52,20 @@ matrix:
...
         - codecov
+    - stage: deploy
+      python: 3.4
+      install:
+        - pip3 install --upgrade pip
+        - pip3 install twine
+      before_script: skip
+      script:
+        - python3 setup.py bdist_wheel
+      deploy: *_deploy
```
vmarkovtsev: Don't offend Python 3.5 and 3.6

.travis.yml
```diff
@@ -12,6 +12,10 @@ addons:
       - libboost-all-dev
       - libxml2-dev
       - libsnappy-dev
+stages:
```
vmarkovtsev: I have a feeling that we can avoid stages (they are slow in case of long setup like we have here) and implement it as a regular `deploy` in our Python tasks

.travis.yml
zurk: But we have `sourced_ml-0.4.4-py3-none-any.whl` as output, so it should work for any python3. And that the reason why we have only one wheel for python3.4,5,6 and need to use stages (your second comment.).

Correct me, if I am wrong.

vmarkovtsev: Right, there are no native extensions. Regarding stages, you can restrict deployment to specific python version, so stages are not required.

.travis.yml
zurk: https://docs.travis-ci.com/user/conditional-builds-stages-jobs/#Conditional-Builds

vmarkovtsev: Nope, this won't do. You need to explain it with normal words.

.travis.yml
zurk: but we first need to test our code for all pythons and only after that deploy. 

if we restrict deployment to specific python version, we can publish wheel that will fail on another pythons.

vmarkovtsev: Fair enough again since we do not build binary packages. My desperate attempt to speed up CI failed

.travis.yml
zurk: Anyway, I add a comment, see new commit below.

PR #194, opened 2018-03-14
Fix locales in Dockerfile

zurk:
We need to reconfigure locales. It was `C.UTF-8` (from srcd/spark:2.2.0_v2 image or its parent) and we need `en_US.UTF-8`.

To avoid such behavior:
```
docker run -it srcd/ml-core bash
bash: warning: setlocale: LC_ALL: cannot change locale (en_US.UTF-8)
root@491e69768462:/# env
HOSTNAME=491e69768462
TERM=xterm
spark_ver=2.2.0
LC_ALL=en_US.UTF-8
CA_CERTIFICATES_JAVA_VERSION=20140324
PYSPARK_PYTHON=python3
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/opt/spark/bin
PWD=/
hadoop_ver=2.7.3
JAVA_HOME=/usr/lib/jvm/java-8-openjdk-amd64
LANG=C.UTF-8
JAVA_VERSION=8u111
SHLVL=1
HOME=/root
JAVA_DEBIAN_VERSION=8u111-b14-2~bpo8+1
BROWSER=/browser
PYSPARK_PYTHON_DRIVER=python3
_=/usr/bin/env
```

### DIFF
--- a/Dockerfile.core
+++ b/Dockerfile.core
@@ -22,7 +22,10 @@ echo\n\' > /browser && \
     chmod +x /browser
 
 ENV BROWSER /browser
+RUN echo "en_US.UTF-8 UTF-8" >> /etc/locale.gen && \
+    dpkg-reconfigure -f noninteractive locales
 ENV LC_ALL en_US.UTF-8
+ENV LANG en_US.UTF-8
 ENV PYSPARK_PYTHON python3
 ENV PYSPARK_PYTHON_DRIVER python3
 


Dockerfile.core
```diff
@@ -22,7 +22,10 @@ echo\n\' > /browser && \
     chmod +x /browser
 
 ENV BROWSER /browser
+RUN echo "en_US.UTF-8 UTF-8" >> /etc/locale.gen && \
```
vmarkovtsev: Are you sure we need this? Supposedly defining `LANG` should be enough. Please test

zurk: yes, unfortunately, we do. 
if I remove this (`"en_US.UTF-8 UTF-8" >> /etc/locale.gen`) step and run `dpkg-reconfigure -f noninteractive locales` I have
```
perl: warning: Setting locale failed.
perl: warning: Please check that your locale settings:
	LANGUAGE = (unset),
	LC_ALL = "en_US.UTF-8",
	LANG = "en_US.UTF-8"
    are supported and installed on your system.
perl: warning: Falling back to the standard locale ("C").
locale: Cannot set LC_CTYPE to default locale: No such file or directory
locale: Cannot set LC_MESSAGES to default locale: No such file or directory
locale: Cannot set LC_ALL to default locale: No such file or directory
Generating locales (this might take a while)...
Generation complete.
```

and the problem remains. 
The same If I set `LANGUAGE = "en_US.UTF-8"`.

I take the solution from different comments in `https://github.com/tianon/docker-brew-debian/issues/45#issue-157139109`, but I need to find a right way to compose it.

vmarkovtsev: We already run `locale-gen` a few lines above and it does exactly that. I bet that this is a Debian glitch but we need to find the root cause precisely.

zurk: I can delete run of `locale-gen` because it seems to fail and I was not able to solve our problem using this command.

PR #196, opened 2018-03-16
Add sourced-engine version auto-update

zurk:
Anyway, we should use the same versions of `sourced-engine` jar file and python module. So I add this logic to avoid mistakes. 

for now it is 0.5.3.

One thing, should we add `pip` to requirements?

### DIFF
--- a/sourced/ml/utils/engine.py
+++ b/sourced/ml/utils/engine.py
@@ -1,5 +1,6 @@
 import functools
 import logging
+import pip
 
 from sourced.engine import Engine
 from sourced.ml.utils import add_spark_args, create_spark, assemble_spark_config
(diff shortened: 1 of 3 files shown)


sourced/ml/utils/engine.py
```diff
@@ -1,6 +1,8 @@
 import functools
 import logging
+import pip
 
+import sourced.engine
```
vmarkovtsev: looks like not used

sourced/ml/utils/engine.py
```diff
@@ -24,7 +26,7 @@ class EngineDefault:
     Default arguments for create_engine function and __main__
     """
     BBLFSH = "localhost"
-    VERSION = "0.5.1"
+    VERSION = {pkg.key: pkg.version for pkg in pip.get_installed_distributions()}["sourced-engine"]
```
vmarkovtsev: clever

zurk:
@vmarkovtsev fixed, please merge. better to have it in a new souced-ml release.

vmarkovtsev:
CI is broken for a nice obscure reason, have fun!

zurk:
Fixed. Pip v9.0.2 backward incompatible... 
https://github.com/pypa/pip/issues/5081

PR #201, opened 2018-03-22
Added flag to zip and add to spark context ml and engine

r0mainK:
Solves this:  https://github.com/src-d/backlog/issues/1226

First write is to create base directory in the archive, if not included the package seems to not be found, didn't find any other way to do it.


sourced/ml/utils/spark.py
```diff
@@ -68,6 +74,17 @@ def create_spark(session_name,
     session.sparkContext.setLogLevel(spark_log_level)
     # Hide py4j verbose logging (It appears in travis mostly)
     logging.getLogger("py4j").setLevel(logging.WARNING)
+    if dep_zip:
+        tmp_dir = mkdtemp()
```
vmarkovtsev: Please use `~/.cache/source{d}/ml` instead and check if the archive exists. If it exists, fall through to the final `addPyFile`

sourced/ml/utils/spark.py
```diff
@@ -68,6 +73,18 @@ def create_spark(session_name,
     session.sparkContext.setLogLevel(spark_log_level)
     # Hide py4j verbose logging (It appears in travis mostly)
     logging.getLogger("py4j").setLevel(logging.WARNING)
+    if dep_zip:
+        zip_path = os.path.expanduser('~/.cache/source{d}/ml/sourced.zip')
+        if not(os.path.exists(zip_path)):
```
vmarkovtsev: Braces () are not necessary and thus should be deleted

sourced/ml/utils/spark.py
```diff
@@ -68,6 +73,18 @@ def create_spark(session_name,
     session.sparkContext.setLogLevel(spark_log_level)
     # Hide py4j verbose logging (It appears in travis mostly)
     logging.getLogger("py4j").setLevel(logging.WARNING)
+    if dep_zip:
+        zip_path = os.path.expanduser('~/.cache/source{d}/ml/sourced.zip')
```
vmarkovtsev: We use double quotes in our projects, see https://github.com/src-d/guide/blob/master/engineering/python-style-guide.md Rationale: relax transitions from Go to Python and back

PR #204, opened 2018-03-22
Improve token parser

zurk:
- Add small subtokens not only to the next token but also to the previous:
before:
`'sourced.ml.algorithms.uast_ids_to_bag' -> ['sourc', 'algorithm', 'mlalgorithm', 'uast', 'ids', 'bag', 'tobag']`
after: 
`'sourced.ml.algorithms.uast_ids_to_bag' -> ["sourc", "sourcedml", "algorithm", "mlalgorithm", "uast", "ids", "idsto", "bag", "tobag"]`
- Add test cases
- refactor `prev_p[0]` to `ret.prev_p`

### DIFF
--- a/sourced/ml/tests/test_token_parser.py
+++ b/sourced/ml/tests/test_token_parser.py
@@ -9,19 +9,30 @@ class TokenParserTests(unittest.TestCase):
         self.tp = TokenParser(stem_threshold=4, max_token_length=20)
 
     def test_process_token(self):
-        self.assertEqual(
-            list(self.tp.split("set /for. *&PrintAll")),
-            ["set", "for", "print", "all"])
-        self.assertEqual(
-            list(self.tp.split("JumpDown not Here")),
-            ["jump", "down", "not", "here"])
+
         _max_token_length = self.tp.max_token_length
         self.tp.max_token_length = 100
-        self.assertEqual(list(self.tp.process_token("sourced.ml.algorithms.uast_ids_to_bag")),
-                         ["sourc", "sourcedml", "algorithm", "mlalgorithm",
-                          "uast", "ids", "idsto", "bag", "tobag"])
-        self.assertEqual(list(self.tp.process_token("a.b.c.d")), ["a", "b", "c", "d"])
 
+        tokens = [
+            ("sourced.ml.algorithms.uast_ids_to_bag",
+             ["sourc", "sourcedml", "algorithm", "mlalgorithm",
+              "uast", "ids", "idsto", "bag", "tobag"]),
+            # Bad example. Parser failed to parse it correctly
+            ("WORSTnameYOUcanIMAGINE", ['worst', 'name', 'nameyo', 'ucan', 'youcan', 'imagin']),
+            # Another bad example. Parser failed to parse it correctly
+            ("ONE_M0re_.__badId.example", ['one', 'onem', 're', 'bad', 'rebad',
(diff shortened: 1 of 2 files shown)


sourced/ml/tests/test_token_parser.py
```diff
@@ -24,7 +32,8 @@ def test_split(self):
         self.assertEqual(list(self.tp.split("PrintAllExcept")), ["print", "all", "except"])
         self.assertEqual(
             list(self.tp.split("print really long line")),
-            ["print", "really", "long"])
+            # 'longli' is expected artifact due to edge effects
+            ["print", "really", "long", "longli"])
```
vmarkovtsev: Please add more tests

vmarkovtsev:
@zurk ping

zurk:
What kind of tests do you mean?
Just more examples?

vmarkovtsev:
Yes, all sorts of the craziest identifiers you can imagine.

sourced/ml/tests/test_token_parser.py
```diff
@@ -9,19 +9,30 @@ def setUp(self):
...
-                         ["sourc", "sourcedml", "algorithm", "mlalgorithm",
-                          "uast", "ids", "idsto", "bag", "tobag"])
-        self.assertEqual(list(self.tp.process_token("a.b.c.d")), ["a", "b", "c", "d"])
 
+        tokens = [
+            ("sourced.ml.algorithms.uast_ids_to_bag",
+             ["sourc", "sourcedml", "algorithm", "mlalgorithm",
+              "uast", "ids", "idsto", "bag", "tobag"]),
+            # Bad example. Parser failed to parse it correctly
+            ("WORSTnameYOUcanIMAGINE", ['worst', 'name', 'nameyo', 'ucan', 'youcan', 'imagin']),
```
vmarkovtsev: This is too bad ^, need to fix the parser on this example.

vmarkovtsev:
@zurk Ping here, I want this change because it affects Waren's splitter

zurk:
ok, Maybe, it is not a good time to fix this example? because also I could use imagine something like:
`WORSTNameYOUCanIMAGINE` and everything collides... 

My suggestion leave it as is for now.

vmarkovtsev:
All right, please promise to finish this after 15th :smile:

zurk:
I promise :)
Do You leave PR open for now?

vmarkovtsev:
Yes.

zurk:
So, @vmarkovtsev what should we do with the following two examples?
`
WORSTNameYOUCanIMAGINE
`
and (watch for caps)
`
WORSTnameYOUcanIMAGINE
`
should one of this be considered as not so mad and be fixed?

right now we have next splitting:
`WORSTNameYOUCanIMAGINE` to `['worst', 'name', 'ucan', 'imagin']`
`WORSTnameYOUcanIMAGINE` to `['worstn', 'ame', 'you', 'can', 'imagin']`

My idea that they are equal and fix for one broke another and vice versa.

Nerual splitter solves all this problem :)

vmarkovtsev:
Now that we have our neural splitter, it is much easier indeed. Still, I expect

```
"WORSTnameYOUcanIMAGINE": ["worst", "name", "you", "can", "imagine"]
```

with a few alternative chunks at worst.

zurk:
Now it should work

PR #206, opened 2018-03-24
Fix bug in language mapping and improve code readability

warenlg:

### DIFF
--- a/sourced/ml/transformers/content2ids.py
+++ b/sourced/ml/transformers/content2ids.py
@@ -102,7 +102,6 @@ class IdentifiersToDataset(Transformer):
         self.idfreq = idfreq
 
     def __call__(self, rows: RDD):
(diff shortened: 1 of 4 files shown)


sourced/ml/transformers/content2ids.py
```diff
@@ -96,43 +95,58 @@ def build_mapping():
...
+                    num_repos=x[1][0],
+                    num_files=x[1][1],
+                    num_occ=x[1][2]))
+
+    def process(self, rows: RDD):
+        """
+        Process rows to gather identifier without frequencies.
+        """
+        return rows \
+            .map(lambda x: x[0]) \
```
smola: You might use `.keys()` instead of `map(...)` here.

warenlg: True, thanks for your review!

PR #208, opened 2018-03-27
Fix processing failure if bblfsh unable to deserialize response

zurk:
If an error happens in `self.parse_uast` it should not fail all computation.

Related to https://github.com/bblfsh/client-python/issues/89

### DIFF
--- a/sourced/ml/transformers/basic.py
+++ b/sourced/ml/transformers/basic.py
@@ -5,6 +5,7 @@ from pyspark import StorageLevel, Row, RDD
 from pyspark.sql import DataFrame
 
 from sourced.ml.transformers.transformer import Transformer
+from sourced.ml.transformers.uast2bag_features import Uast2BagFeatures
 from sourced.ml.utils import EngineConstants, assemble_spark_config, create_spark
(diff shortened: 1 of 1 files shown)


smola:
We should investigate why are we getting invalid protobuf messages at that point in the code. If there was some error when making a request to Babelfish, we could retry a number of times, but at this point it's too late.

sourced/ml/transformers/basic.py
```diff
@@ -182,6 +183,11 @@ def deserialize_uast(self, row: Row):
         if not row[EngineConstants.Columns.Uast]:
             return
         row_dict = row.asDict()
-        row_dict[EngineConstants.Columns.Uast] = [
-            self.parse_uast(uast) for uast in row[EngineConstants.Columns.Uast]]
+        try:
```
vmarkovtsev: This is suboptimal: we lose everything if a single UAST request fails. I would rather code a `for` loop and check each `parse_uast` independently with `continue`

zurk: done.

Issue #210, opened 2018-03-28
repos2bow fails if --docfreq points to a file in non-existing directory


smola:
Running: `srcml repos2bow --repositories ~/dev/working/siva-good --bow data/BOW --feature lit --languages Java  --docfreq data/docfreq --persist MEMORY_AND_DISK`

Directory `data` did not exist. The job failed pretty late with:

```
INFO:repos2bow:Writing docfreq to data/docfreq                                  
INFO:docfreq:Ordering the keys...
Traceback (most recent call last):
  File "ENV/bin/srcml", line 11, in <module>
    load_entry_point('sourced-ml', 'console_scripts', 'srcml')()
  File "/home/smola/dev/go/src/github.com/src-d/ml/sourced/ml/__main__.py", line 187, in main
    return handler(args)
  File "/home/smola/dev/go/src/github.com/src-d/ml/sourced/ml/cmd_entries/repos2bow.py", line 80, in repos2bow_entry
    return repos2bow_entry_template(args)
  File "/home/smola/dev/go/src/github.com/src-d/ml/sourced/ml/utils/engine.py", line 73, in wrapped_pause
    return func(cmdline_args, *args, **kwargs)
  File "/home/smola/dev/go/src/github.com/src-d/ml/sourced/ml/cmd_entries/repos2bow.py", line 68, in repos2bow_entry_template
    .save(args.docfreq)
  File "/home/smola/dev/go/src/github.com/src-d/ml/ENV/lib/python3.6/site-packages/modelforge/model.py", line 269, in save
    self._write_tree(tree, output)
  File "/home/smola/dev/go/src/github.com/src-d/ml/ENV/lib/python3.6/site-packages/modelforge/model.py", line 285, in _write_tree
    asdf.AsdfFile(final_tree).write_to(output, all_array_compression=ARRAY_COMPRESSION)
  File "/home/smola/dev/go/src/github.com/src-d/ml/ENV/lib/python3.6/site-packages/asdf/asdf.py", line 890, in write_to
    with generic_io.get_file(fd, mode='w') as fd:
  File "/home/smola/dev/go/src/github.com/src-d/ml/ENV/lib/python3.6/site-packages/asdf/generic_io.py", line 1186, in get_file
    fd = atomicfile.atomic_open(realpath, realmode)
  File "/home/smola/dev/go/src/github.com/src-d/ml/ENV/lib/python3.6/site-packages/asdf/extern/atomicfile.py", line 139, in atomic_open
    delete=False)
[...]

PR #211, opened 2018-03-28
Create output directory in repos2bow if it does not exists

zurk:
Fix https://github.com/src-d/ml/issues/210

### DIFF
--- a/sourced/ml/cmd_entries/repos2bow.py
+++ b/sourced/ml/cmd_entries/repos2bow.py
@@ -1,5 +1,6 @@
 import argparse
 import logging
+import os
 from uuid import uuid4
 
 from sourced.ml.extractors import create_extractors_from_args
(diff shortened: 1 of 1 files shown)


sourced/ml/cmd_entries/repos2bow.py
```diff
@@ -61,6 +62,12 @@ def repos2bow_entry_template(args, select=HeadFiles, cache_hook=None):
     log.info("Calculating the document frequencies...")
     df = uast_extractor.link(BagFeatures2DocFreq()).execute()
     log.info("Writing docfreq to %s", args.docfreq)
+    dirs = os.path.split(args.docfreq)[0]
```
vmarkovtsev: You should put this inside Modelforge's `Model.save`; add the new argument `create_missing_dirs` which is `True` by default.

PR #216, opened 2018-04-03
Adding hook to repos2bow_template

r0mainK:
Added a second hook to this template, in order to be able to save bags to DB. Will be exploited by apollo the same way the `cache_hook`is used to save the metadata to DB


sourced/ml/cmd_entries/repos2bow.py
```diff
@@ -66,12 +66,15 @@ def repos2bow_entry_template(args, select=HeadFiles, cache_hook=None):
...
         .greatest(args.vocabulary_size) \
         .save(args.docfreq)
-    uast_extractor \
+    bags_writer= uast_extractor \
         .link(BagFeatures2TermFreq()) \
         .link(TFIDF(df_model)) \
         .link(document_indexer) \
-        .link(Indexer(Uast2BagFeatures.Columns.token, df_model.order)) \
-        .link(BOWWriter(document_indexer, df_model, args.bow, args.batch)) \
+        .link(Indexer(Uast2BagFeatures.Columns.token, df_model.order))\
```
vmarkovtsev: no space before \

sourced/ml/cmd_entries/repos2bow.py
```diff
@@ -66,12 +66,15 @@ def repos2bow_entry_template(args, select=HeadFiles, cache_hook=None):
...
+        .link(Cacher.maybe(args.persist))
```
vmarkovtsev: I have a feeling that `save_hook` should allow to change BOWWriter to an arbitrary transformer. Thus you do not need to cache. I don't know why it is needed upstream; another option is to filter the data through save_hook() and link everything together without a fork.

r0mainK:
@vmarkovtsev okay now piping through the hook before writing the model, took out the cach. I had to modify  a bit the class in apollo in order for it to work

vmarkovtsev:
@r0mainK Please add signed-off-by next time. It is not strictly required for source{d} employees but is fair to the external contributors.

PR #224, opened 2018-04-06
Minor pipeline_graph imrovement

zurk:
To be compatible with current apollo


vmarkovtsev:
@zurk This is a hack. Remove this pipeline function call from those places where `--graph` is not present.

PR #225, opened 2018-04-09
New default parameters

EgorBu:
Add new default parameters to extractors/algorithms based on hyperoptimization experiments

### DIFF
--- a/sourced/ml/algorithms/uast_inttypes_to_nodes.py
+++ b/sourced/ml/algorithms/uast_inttypes_to_nodes.py
@@ -11,7 +11,7 @@ class Uast2QuantizedChildren(Uast2BagThroughSingleScan):
     Converts a UAST to a bag of children counts.
     """
 
-    def __init__(self, npartitions: int=10):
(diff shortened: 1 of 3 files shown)


sourced/ml/cmd_entries/args.py
```diff
@@ -18,9 +18,9 @@ def _get_help_string(self, action):
 
 def add_split_stem_arg(my_parser: argparse.ArgumentParser):
     my_parser.add_argument(
-        "--split", action="store_true",
-        help="Split identifiers based on special characters or case changes. For example split "
-             "'ThisIs_token' to ['this', 'is', 'token'].")
+        "--split", action="store_false",
```
zurk: we should change the name to `--no-split`

EgorBu: yes, `split or not to split - that's the question` - naming became confusing in this case)

vmarkovtsev:
@EgorBu Still waiting for `--no-split`

PR #228, opened 2018-04-11
Add repartition into repo2coocc

zurk:
It helps a lot with performance issues.
I add this transformer to all other subprograms later.

### DIFF
--- a/sourced/ml/cmd_entries/args.py
+++ b/sourced/ml/cmd_entries/args.py
@@ -17,6 +17,16 @@ class ArgumentDefaultsHelpFormatterNoNone(argparse.ArgumentDefaultsHelpFormatter
         return super()._get_help_string(action)
 
 
+def add_repartitioner_arg(my_parser: argparse.ArgumentParser):
+    my_parser.add_argument(
+        "--partitions", required=False, default=None, type=int,
+        help="Performs data repartition to specified number of partitions. " 
+             "Nothing happens if parameter is unset.")
+    my_parser.add_argument(
+        "--shuffle", action="store_true",
+        help="Use repartition instead of coalesce.")
+
+
 def add_split_stem_arg(my_parser: argparse.ArgumentParser):
     my_parser.add_argument(
         "--split", action="store_true",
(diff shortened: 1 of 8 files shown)


sourced/ml/transformers/basic.py
```diff
@@ -62,6 +62,22 @@ def __call__(self, head: RDD):
         return head
 
 
+class Repartitioner(Transformer):
+    def __init__(self, partitions: int, **kwargs):
+        super().__init__(**kwargs)
+        self.partitions = partitions
+
+    def __call__(self, head: RDD):
+        return head.repartition(self.partitions)
```
vmarkovtsev: How about `coalesce()`? I would add a flag to `__init__` and do `coalesce()` by default.

zurk: repartition is actually 
```python
return self.coalesce(numPartitions, shuffle=True)
```
So we can add `shuffle` flag, yes, but I think in case of performance it can be not so good idea because what we saw is really skewed partitions after reading siva files. 

I can run some performance tests locally if you want.

zurk: after repartition, I can see that last jobs finish fast one after another and we do not have a problem waiting while one process will finish.

zurk: about skewed partitions, see the image in 
https://github.com/src-d/engine/issues/351

vmarkovtsev: This may be not good for your specific use case but good for the others, this transformer is universal. Feel free to hardcode `shuffle=True` in repos2coocc but make it false by default in the transformer.

zurk:
There was no additional comment, so it was just rebased on top of current master.

sourced/ml/cmd_entries/args.py
```diff
@@ -17,6 +17,16 @@ def _get_help_string(self, action):
         return super()._get_help_string(action)
 
 
+def add_repartitioner_arg(my_parser: argparse.ArgumentParser):
+    my_parser.add_argument(
+        "--partitions", required=False, default=None, type=int,
```
vmarkovtsev: Performs data repartitioning

Use RDD.repartition() instead of RDD.coalesce()

PR #223, opened 2018-04-05
Parquet loader

r0mainK:
Based of [PR 222](https://github.com/src-d/ml/pull/222), should not be merged before it

- Added language selection to `repo2id`. Since it already had transformers a bit redundant with `UastExtractor` I split it, in order to get rid of `ContentExtractor` class
- Modified the parsers (was a bit of a mess so i rearranged them to have as little as possible). Among other things i put a default language flag (to all), put pause flag in `add_spark`, modified `add_repo` and `add_features`
- Added parquet loader, pause and graph to all `repo2` commands

PR #229, opened 2018-04-16
Move pause arg to spark parser, add pause to repos2id

r0mainK:
First part of split of [old PR](https://github.com/src-d/ml/pull/223/files), moving pause to the spark args parser. It seems more logical, given we will only use in conjunction with spark.

This also takes out pip of imports: since version 10, we need new way to get engine version. Now using `pkg_ressources` to do it

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,4 +8,3 @@ numpy>=1.14
 humanize>=0.5.0,<0.6
 pyspark>=2.2.0,<2.2.1
 pygments>=2.2.0,<3.0
-pip!=9.0.2
--- a/setup.py
+++ b/setup.py
@@ -34,8 +34,7 @@ setup(
                       "pyspark>=2.2.0,<2.2.1",
                       "humanize>=0.5.0",
                       "parquet>=1.2,<2.0",
-                      "pygments>=2.2.0,<3.0",
-                      "pip!=9.0.2"] + typing,
+                      "pygments>=2.2.0,<3.0"] + typing,
     extras_require={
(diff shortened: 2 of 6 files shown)


sourced/ml/utils/__init__.py
```diff
@@ -1,6 +1,6 @@
 from sourced.ml.utils.bigartm import install_bigartm
-from sourced.ml.utils.bblfsh_roles import IDENTIFIER, QUALIFIED, LITERAL
-from sourced.ml.utils.spark import add_spark_args, create_spark, assemble_spark_config, \
+from sourced.ml.utils.bblfsh_roles import IDENTIFIER, LITERAL, QUALIFIED
```
vmarkovtsev: This is the same, only the order has changed, please revert

r0mainK: done

sourced/ml/utils/__init__.py
```diff
@@ -1,6 +1,6 @@
 from sourced.ml.utils.bigartm import install_bigartm
 from sourced.ml.utils.bblfsh_roles import IDENTIFIER, QUALIFIED, LITERAL
-from sourced.ml.utils.spark import add_spark_args, create_spark, assemble_spark_config, \
+from sourced.ml.utils.spark import add_spark_args, assemble_spark_config, create_spark, \
```
vmarkovtsev: Same here

r0mainK: done

PR #220, opened 2018-04-03
Add code for identifier, role dataset generation

zurk:
This PR is based on https://github.com/src-d/ml/pull/218

Includes:
- new algorithm
- new extractor
- new cmd entry point

PR #230, opened 2018-04-16
Split UAST extractor and add parquet loader

r0mainK:
**AFTER REBASING:**

- First commit is to split `UastExtractor` in order to separate the language selection from it. This allows us to remove part of the `ContentExtractor` class, add a `blacklist` feature to the selector (exclude languages instead of include them), and overall simplify this transformer. We may want to add `blacklist` as a flag ? Currently it is only used internally by `repos2ids`in order to select all non-null languages, but we could imagine a user wanting to use `repos2ids` to extract by language, or a user wanting to select languages by simply excluding some, e.g. all languages but `Java`

- Second commit replaces in all `repo2` commands the old loader with `create_uast_source`, so we can use parquet files as input always. I think it could be a good idea to update docs as well, so the uesr knows the parquet are expected to have been preprocessed in the apollo sense, i.e. filtered by language and uast already extracted. I aslo added the `pipeline_graph` to all commands that did not have it --no sense in not doing so

- Third commit adds testing for the parquet loader, as well as a dummy test for projector. Was wandering whether we should add a test for engine now, after the discussion we had


r0mainK:
Rebased off master to integrate `PR229`

sourced/ml/cmd_entries/args.py
```diff
@@ -67,3 +67,13 @@ def add_feature_args(my_parser: argparse.ArgumentParser, required=True):
...
+
+
+def add_bow_args(my_parser: argparse.ArgumentParser):
+    my_parser.add_argument(
+        "--bow", required=True, help="[OUT] The path to the Bag-Of-Words model.")
+    my_parser.add_argument(
+        "--batch", default=BOWWriter.DEFAULT_CHUNK_SIZE, type=int,
+        help="The maximum size of a single BOW file in bytes.")
+    my_parser.add_argument(
+        "--parquet", action="store_true", help="If it's parquet input.")
```
vmarkovtsev: Please rephrase the help message

vmarkovtsev: I don't see any relation between parquet and bow here...

r0mainK: there is none really, this is only added here until https://github.com/src-d/ml/pull/231 and https://github.com/src-d/ml/pull/232 are merged for consistency (232 changes the parsers in a simpler way). anyway, changing help msg to "Use Parquet files as input."

r0mainK:
@vmarkovtsev so regarding the idea of having a transformer pipeline grouping as many common tasks as possible, Konst actually told me the same thing during the draft of this PR, but slightly differently. Since we add the possibility to load repos from Parquet files for all `repo2 ...` tasks, he told me to create this function: https://github.com/src-d/ml/pull/231/files#diff-370750368c8314a487dfbae4ca854901R227 which achieves the same idea of chained Transformers. I hadn't included in this PR to makes reviews easier (but maybe splitting in 4 parts was a bit much). anyway giving that should I still go ahead with a `UASTPipeline` or ?

sourced/ml/transformers/basic.py
```diff
@@ -136,9 +136,16 @@ def __call__(self, files: DataFrame) -> DataFrame:
...
         filtered_by_lang = classified.filter(lang_filter)
+        return filtered_by_lang
+
+
+class UastExtractor(Transformer):
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+
+    def __call__(self, files: DataFrame) -> DataFrame:
         from pyspark.sql import functions
```
zurk: BTW, why do we have this import here?

sourced/ml/transformers/content2ids.py
```diff
@@ -146,17 +146,3 @@ def process(self, rows: RDD):
             .map(lambda x: Row(
                     token=x,
                     token_split=" ".join(TokenParser(min_split_length=1).split(x))))
-
-
-class ContentExtractor(Transformer):
```
zurk: you change it to
```python
-        .link(ContentExtractor()) \
+        .link(LanguageSelector(languages=args.languages)) \
```
here: https://github.com/src-d/ml/pull/230/files#diff-0a78c1a1a739541ac75c64d15c310dbeL15
and it is not the same.

We filter here only by the rule `"lang is not null"` and in `LanguageSelector` we have different. 
Also, we filter by content. 
Please check. 
You can
1.  leave it as is or 
2. modify `LanguageSelector` and maybe add some additional transformers.

r0mainK: 1. Regarding content, your are right I forgot to add the filtering to the ContentExtractor
2. Regarding language, after speaking with Waren he told me that yes, we could not filter language that way. Whenever Babelfish supports as much languages as Pygments it will not be a problem, however until then we should keep the "lang is not null" rule. I think the most logical is to do the following:
- make default language param to None
- if None, use rule "lang is not null"
- else do it that way
gonna push change, tell me what u think

zurk: > if language is None, use rule "lang is not null"
it is counterintuitive a bit. You can pass it as a string "lang is not null" instead of None.

r0mainK: okay, but I dont want to hardcode a test `if self.languages == lang is not null`, and using `lang_filter = classified.lang == "not null"` or `lang_filter = classified.lang == "lang is not null"` will not work. I think in this case it is best to add a boolean `blacklist` that defaults to False, but if set to True will exclude by language, e.g. for languages = [null, Python, Java]: 
- `blacklist` is False: `lang_filter = lang == null OR lang == Python OR lang == Java`
- `blacklist` is True: `lang_filter = lang != null AND lang != Python AND lang == Java`
That way we will be able to exclude languages in an easier way then by passing a long list of languages to keep

zurk: ok.

vmarkovtsev:
Waiting for https://github.com/src-d/ml/pull/240 to remove some of the edits from here

r0mainK:
@vmarkovtsev @zurk finished rebasing off #240, I closed #232 and #231 since all changes they brought are either merged, or in this PR (thought it would be easier to review this way). I updated the PR description so you can have an overview of the _new_ new changes

vmarkovtsev:
@zurk Is on vacation. @EgorBu do you see any problems?

sourced/ml/cmd_entries/repos2coocc.py
```diff
@@ -3,24 +3,20 @@
...
+    BagFeatures2DocFreq, Uast2BagFeatures, Counter, UastRow2Document, create_uast_source
 from sourced.ml.utils.engine import pipeline_graph, pause
 
 
 @pause
 def repos2coocc_entry(args):
     log = logging.getLogger("repos2coocc")
-    engine = create_engine("repos2coocc-%s" % uuid4(), **args.__dict__)
     id_extractor = IdentifiersBagExtractor(docfreq_threshold=args.min_docfreq,
                                            split_stem=args.split_stem)
```
EgorBu: There is no argument `--split-stem`. https://github.com/src-d/ml/pull/220/files - this is the PR that broke cooc extraction because `split-stem` was replaced with `split` and in `repos2coocc_entry` it's still `split_stem=args.split_stem`

sourced/ml/transformers/basic.py
```diff
@@ -225,13 +237,18 @@ def create_parquet_loader(session_name, repositories,
     return parquet
 
 
-def create_uast_source(args, session_name, select=HeadFiles):
+def create_uast_source(args, session_name, select=HeadFiles, language_selector=None,
+                       extract_uast=True):
```
EgorBu: Please add docs. It's not clear why it should return `root` and `start_point`. 
`extract_uast` and other arguments are not used in case of `parquet` - maybe it's better to split this function.
It depends on `add_repo2_args` where `parquet` flag is added - so it should be checked that `args` has `parquet` or it should be mentioned that `add_repo2_args` should be used on a parser.

sourced/ml/transformers/basic.py
```diff
@@ -225,13 +237,18 @@ def create_parquet_loader(session_name, repositories,
...
     if args.parquet:
         start_point = create_parquet_loader(session_name, **args.__dict__)
         root = start_point
```
EgorBu: I think it should check the availability of `uast`  field according to argument `extract_uast`.
You can receive a list of column names like this `root.schema.names` - it will allow debugging errors faster and easier.

sourced/ml/transformers/basic.py
```diff
@@ -125,21 +125,33 @@ def __call__(self, head: RDD):
...
-        return uasts
+        return filtered_by_lang
+
+
+class UastExtractor(Transformer):
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+
+    def __call__(self, files: DataFrame) -> DataFrame:
+        return files.extract_uasts().where(functions.size(functions.col("uast")) > 0)
```
EgorBu: good to add an explanation why we need only uasts longer than 0. (if uast was extracted -> not empty list of results)

sourced/ml/cmd_entries/repos2roles_and_ids.py
```diff
@@ -1,24 +1,25 @@
...
-        .link(UastExtractor(languages=args.languages)) \
+    start_point \
         .link(UastRow2Document()) \
         .link(Cacher.maybe(args.persist)) \
         .link(UastDeserializer()) \
         .link(Uast2BagFeatures([RolesAndIdsExtractor(args.split)])) \
         .link(Rower(lambda x: dict(identifier=x[0][0], role=x[1]))) \
         .link(CsvSaver(args.output)) \
         .execute()
+    pipeline_graph(args, log, root)
```
EgorBu: is it the only reason to have `root` in `create_uast_source`?

sourced/ml/transformers/basic.py
```diff
@@ -225,13 +237,18 @@ def create_parquet_loader(session_name, repositories,
...
         root = create_engine(session_name, **args.__dict__)
+        if language_selector is None:
+            language_selector = LanguageSelector(languages=args.languages)
         start_point = Ignition(root, explain=args.explain) \
             .link(select()) \
-            .link(UastExtractor(languages=args.languages))
+            .link(language_selector)
+        if extract_uast:
+            start_point = start_point.link(UastExtractor())
     return root, start_point
```
EgorBu: I think it could be worth to create some hook and pass `pipeline_graph` instead of returning `root`

sourced/ml/cmd_entries/repos2coocc.py
```diff
@@ -40,11 +36,11 @@ def repos2coocc_entry(args):
         .greatest(args.vocabulary_size) \
         .save(args.docfreq)
 
-    token2index = engine.session.sparkContext.broadcast(df_model.order)
+    token2index = root.session.sparkContext.broadcast(df_model.order)
```
EgorBu: Can you use `session` from `start_point`?

sourced/ml/cmd_entries/repos2coocc.py
r0mainK: okay, changing to split

vmarkovtsev:
@EgorBu do you approve?

vmarkovtsev:
@r0mainK better late than never! Good job.

PR #234, opened 2018-04-17
Doc: re-using Apache Spark installation

### LINKED ISSUE #233: Option to avoid downloading Apache Spark on every install
Sorry if I'm missing something here, but right now `sourced.ml` library has, what seems to be a [mandatory dependency](https://github.com/src-d/ml/blob/master/requirements.txt#L9) on Apache Spark though `pip`.

As a user of src-d/ml, I would expect this library to let me avoid downloading 188Mb of Apache Spark as a dependency, and instead let me choosing my local Apache Spark installation \w pySpark through `$SPARK_HOME`.

Same way as it's not fetching Tensorflow and libxml2 and just mention them in https://github.com/src-d/ml#installation

bzz:
Fixes #233

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,11 +22,20 @@ Here is the list of proof-of-concept projects which are built using sourced.ml:
 
 ## Installation
 
+### With Apache Spark included
 ```
 pip3 install sourced-ml
 ```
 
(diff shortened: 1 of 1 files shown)


vmarkovtsev:
Awesome, thanks.

PR #237, opened 2018-04-19
Fix: Add try/except to VERSION detection

r0mainK:
Simple PR to add a `try`/`except` block in the engine version detection, and default to "0.5.7" if the engine is not detected. This solves a very dumb bug that happens when using Spark in Cluster mode with `ml`, that I describe below. I had not seen it until now because I was replacing this detection with the appropriate `engine` version, e.g. `VERSION = "0.5.7"`, due to a different bug (something to do with `pip` version 9.2). Since with version 10 of `pip` we now use `pkg_ressources` to detect the `engine` version, I updated the detection and the bug popped up.

**Bug description:** 

When using `ml` with Spark in Cluster mode, we zip `ml` and `engine` then add the zip via `addPyFile` to each worker. If this is not done, our tasks fail, given workers do not have `ml` and `engine` installed. This means that neither `pip` nor `pkg_ressources` can detect ml/engine from within the workers, which is problematic because in `sourced/ml/utils/engine.py` that is exactly what we try to do in order to detect the engine version.

Now the annoying thing is that we detect the engine version to specify the correct `engine` jar to be passed to the workers, but once inside the workers there is no need to do so. However, when running jobs this still happens, probably due to some import that leads the worker to check out `sourced/ml/utils/engine.py`, which raises a `DistributionNotFound` error.

This will solve the error by defaulting to a possibly fake engine version, which in all cases will not be used. It is not problematic because if the `engine` is not present, since we still do `from sourced.engine import Engine` in the same file, we will get an `ImportError`.

### DIFF
--- a/sourced/ml/utils/engine.py
+++ b/sourced/ml/utils/engine.py
@@ -1,6 +1,6 @@
 import functools
 import logging
-from pkg_resources import get_distribution
+from pkg_resources import get_distribution, DistributionNotFound
 from sourced.engine import Engine
 from sourced.ml.utils.spark import add_spark_args, assemble_spark_config, create_spark
 
@@ -24,7 +24,10 @@ class EngineDefault:
     Default arguments for create_engine function and __main__
     """
     BBLFSH = "localhost"
-    VERSION = get_distribution("sourced-engine").version
+    try:
+        VERSION = get_distribution("sourced-engine").version
+    except DistributionNotFound:
+        VERSION = "0.5.7"
 
 
 def add_engine_args(my_parser, default_packages=None):
--- a/sourced/ml/utils/engine.py
+++ b/sourced/ml/utils/engine.py
@@ -1,5 +1,6 @@
 import functools
 import logging
+import requests
 from pkg_resources import get_distribution, DistributionNotFound
 from sourced.engine import Engine
 from sourced.ml.utils.spark import add_spark_args, assemble_spark_config, create_spark
@@ -19,22 +20,11 @@ class EngineConstants:
         Uast = "uast"
 
 
-class EngineDefault:
-    """
-    Default arguments for create_engine function and __main__
-    """
-    BBLFSH = "localhost"
-    try:
-        VERSION = get_distribution("sourced-engine").version
-    except DistributionNotFound:
-        VERSION = "0.5.7"
-
-
 def add_engine_args(my_parser, default_packages=None):
     add_spark_args(my_parser, default_packages=default_packages)
-    my_parser.add_argument("--bblfsh", default=EngineDefault.BBLFSH,
+    my_parser.add_argument("--bblfsh", default=None,
                            help="Babelfish server's address.")
-    my_parser.add_argument("--engine", default=EngineDefault.VERSION,
+    my_parser.add_argument("--engine", default=None,
                            help="source{d} engine version.")
     my_parser.add_argument("--repository-format", default="siva",
                            help="Repository storage input format.")
@@ -42,8 +32,9 @@ def add_engine_args(my_parser, default_packages=None):
                            help="Print the PySpark execution plans.")
 
 
-def add_engine_dependencies(engine=EngineDefault.VERSION, config=None, packages=None):
-    config.append("spark.tech.sourced.engine.cleanup.skip=true")
+def add_engine_dependencies(engine, config=None, packages=None):
+    # to clean up unpacked Siva files, see https://github.com/src-d/engine/issues/348
+    config.append("spark.tech.sourced.engine.cleanup.skip=false")
     packages.append("tech.sourced:engine:" + engine)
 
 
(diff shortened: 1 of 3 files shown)


r0mainK:
Since coverage was failing I added tests for `add_bblfsh_dependencies` and `add_engine_dependencies`. I saw that we specified `spark.tech.sourced.engine.cleanup.skip=true`, and changed the value to `false`  because it seems to me we would want to delete the `unpacked` siva files after using them ? 

If not, then will revert, else we can merge this whenever.

r0mainK:
So as explained [here](https://github.com/src-d/engine/issues/348) `spark.tech.sourced.engine.cleanup.skip` should have been set to `false` all along.

sourced/ml/utils/engine.py
```diff
@@ -40,7 +43,7 @@ def add_engine_args(my_parser, default_packages=None):
 
 
 def add_engine_dependencies(engine=EngineDefault.VERSION, config=None, packages=None):
-    config.append("spark.tech.sourced.engine.cleanup.skip=true")
+    config.append("spark.tech.sourced.engine.cleanup.skip=false")
```
vmarkovtsev: But why?

sourced/ml/utils/engine.py
```diff
@@ -24,7 +24,10 @@ class EngineDefault:
     Default arguments for create_engine function and __main__
     """
     BBLFSH = "localhost"
-    VERSION = get_distribution("sourced-engine").version
+    try:
+        VERSION = get_distribution("sourced-engine").version
+    except DistributionNotFound:
+        VERSION = "0.5.7"
```
vmarkovtsev: Would be nice to show a warning in the logs. Also query the version from GitHub API for releases (`import requests`)

sourced/ml/utils/engine.py
vmarkovtsev: OK, now I see. Better to include it into a comment above wit ha link

sourced/ml/utils/engine.py
r0mainK: done

r0mainK:
@vmarkovtsev Still the same coverage error, I had to take out the check on engine version since its subject to change - don't know why but the query to github worked on one side and not the other -> https://travis-ci.org/src-d/ml/jobs/370593538

r0mainK:
Okay so I thought this was good after testing it on k8, but when running apollo today I got this error: `KeyError: 'tag_name'`

After digging a bit, since I was sure the key was correct, I understood why it was happening:
```
>>> j = get("https://api.github.com/repos/src-d/engine/releases/latest") \
...             .json()
>>> j
{'message': "API rate limit exceeded for 88.2.81.45. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)", 'documentation_url': 'https://developer.github.com/v3/#rate-limiting'}
```

Im guessing its because each executor makes a call to the API. Should we revert to hard coding a version ?

smola:
@r0mainK You probably want to make a single API call when starting the driver, then broadcast the version to executors.

r0mainK:
@smola Yeah I see your point, but I think i found a simpler solution - to be honest these defaults are not intended to be used nowhere else then during command parsing as explained at the get go, and the class description, the only reason we need to do all this is bc the class is defined in the outer scope so each executor run the code regardless, probably after some import.

I think the best thing to do is:
- remove the class
- in `add_engine_args` replace defaults with None
- in `add_engine_dependencies` take out the default
- in `create_engine` test if `engine`/`bblfsh`is None, then do the calls.

I just tested it and it works

r0mainK:
@vmarkovtsev if you are satisfied with changes I think you can relaunch the build (need write access to do it on Travis) and this can be merged - Im guessing the build will succeed since it failed only on Python3.6 after not being able to DL Spark

sourced/ml/utils/engine.py
```diff
@@ -1,6 +1,7 @@
 import functools
 import logging
-from pkg_resources import get_distribution
+from requests import get
```
vmarkovtsev: "get" is too common, better to import requests and use `requests.get`

sourced/ml/utils/engine.py
```diff
@@ -24,23 +25,30 @@ class EngineDefault:
...
     """
     BBLFSH = "localhost"
-    VERSION = get_distribution("sourced-engine").version
+    try:
+        VERSION = get_distribution("sourced-engine").version
+    except DistributionNotFound:
+        log = logging.getLogger("engine_version")
+        VERSION = get("https://api.github.com/repos/src-d/engine/releases/latest") \
+            .json()["tag_name"].replace("v", "")
+        log.warning("Engine not found, queried GitHub to get latest release tag (%s)", VERSION)
```
vmarkovtsev: the latest

sourced/ml/utils/engine.py
```diff
@@ -49,10 +57,20 @@ def add_bblfsh_dependencies(bblfsh, config=None):
...
-                  bblfsh=EngineDefault.BBLFSH,
-                  engine=EngineDefault.VERSION,
+                  bblfsh=None,
+                  engine=None,
                   config=None, packages=None, memory="",
                   repository_format="siva", **spark_kwargs):
+    if not bblfsh:
+        bblfsh = "localhost"
+    if not engine:
+        try:
```
vmarkovtsev: This is a duplication, please refactor this into a function and use in two places

r0mainK: my bad I forgot to delete the class `EngineDefaults`, so this is now only used in `create_engine`. the multiple calls to the APIs were done because in the executors there was imports done to this file, which prompted the class to be redefined each time. since the class is used nowhere else I don't think its useful to keep it - I don't really see a situation where we would need to use it either

PR #240, opened 2018-04-30
FIX: Spark loader and Parser

r0mainK:
This PR contains twos fixes, and also changes a bit the requirements file.

1. The first fix concerned the `create_spark` , in order to fix Key errors when loading files I modified the functions signatures, and remove the unused args in the `create_engine` and `create_parquet_loader` functions

2. The second fix concerns the parser. Apollo currently has included changes that are not yet merged in ml, concerning the parser. These were originally to be merged in this PR https://github.com/src-d/ml/pull/232 however in order to get apollo to work I thought it would be best to include them now. This simplifies the parser by a lot, at the moment some arguments will change nothing (e.g. using parquet on all but `repo2bow` will change nothing since the loader is added in another PR) but PR 232 and 230 will change this. To simplify the parser I just looked at all subcommands and looked which had common commands. Here is a summary of all the changes: 

- Move the `add_bow_args` to `cmd_entries/args` -> no reason not to, apollo assumes this already
- Move explain from `add_engine_args` to `add_spark_args` -> given what this flag does it seems logical, we could not want to use the engine but still see the execution plan
- Centralize in `add_repo2_args` all args required by the `repo2` subcommands -> delete `add_extractor_args` and move `--repositories` and `--languages` args in `add_repo2_args`  ; add `--parquet` flag ; add `add_engine_args`
- Centralize in `add_feature` all feature args --> add `--quant` flag
- Create `add_df_args` to centralize all DocFreq related args, and remove them from `add_repo2_args` -> put there `add_vocabulary_size_arg`, `--docfreq` and `--min-docfreq`

3. Finally bump the engine version to < 0.7 and move a bit the imports to avoid error logs during install


requirements.txt
```diff
@@ -1,10 +1,9 @@
...
 bblfsh>=2.2.1,<3.0
 modelforge>=0.5.4-alpha,<0.6
-sourced-engine>=0.5.1,<0.6
+sourced-engine>=0.5.1,<0.7
 parquet>=1.2,<2.0
 numpy>=1.14
 humanize>=0.5.0,<0.6
-pyspark>=2.2.0,<2.2.1
 pygments>=2.2.0,<3.0
+tensorflow>=1.5,<2.0
```
vmarkovtsev: We do not include tensorflow into the requirements: there can be `tensorflow-gpu` too.

requirements.txt
```diff
@@ -1,10 +1,9 @@
 PyStemmer>=1.3,<2.0
-tensorflow>=1.5,<2.0
 bblfsh>=2.2.1,<3.0
 modelforge>=0.5.4-alpha,<0.6
-sourced-engine>=0.5.1,<0.6
+sourced-engine>=0.5.1,<0.7
```
vmarkovtsev: Let's require a specific version as it is not very stable lately

sourced/ml/cmd_entries/args.py
```diff
@@ -29,37 +30,37 @@ def add_vocabulary_size_arg(my_parser: argparse.ArgumentParser):
...
-            # TODO(zurk): get languages from bblfsh directly as soon as
-            # https://github.com/bblfsh/client-scala/issues/68 resolved
-            "Java", "Python", "Go", "JavaScript", "TypeScript", "Ruby", "Bash", "Php"),
+        "--parquet", action="store_true", help="Use Parquet files as input.")
+    my_parser.add_argument(
+        "--graph", help="Write the tree in Graphviz format to this file.")
+    my_parser.add_argument(
+        "-l", "--languages", required=True, nargs="+",
+        choices=("Java", "Python", "Go", "JavaScript", "TypeScript", "Ruby", "Bash", "Php"),
+        default=["Java", "Python", "Go", "JavaScript", "TypeScript", "Ruby", "Bash", "Php"],
```
vmarkovtsev: Konst's TODO still holds. Also please use the same variable twice to avoid duplication

sourced/ml/utils/spark.py
```diff
@@ -29,7 +30,7 @@ def add_spark_args(my_parser, default_packages=None):
         "--config", nargs="+", default=SparkDefault.CONFIG,
         help="Spark configuration (key=value).")
     my_parser.add_argument(
-        "-m", "--memory",
+        "-m", "--memory", default=SparkDefault.MEMORY,
```
vmarkovtsev: This is a bad default, it explodes on any big query. `DISK` + `MEMORY` is better

requirements.txt
r0mainK: okay done

requirements.txt
r0mainK: 0.6.1 has been working for me, I guess we might as well use that one (it also fixes some bug when Siva files are corrupt)

sourced/ml/cmd_entries/args.py
r0mainK: my bad, didnt mean to delete it, on it

r0mainK: also gonna take out the required since it will default to all languages now

r0mainK: my bad, didnt mean to delete it, on it. Im also gonna delte the required=True since we default to all languages now

sourced/ml/utils/spark.py
r0mainK: done

r0mainK: I think you are confusing the `--persist` flag that specifies storage level when caching, and this `--memory` flag that is just a shortcut to define  executor memory, driver memory and driver max result size, and currently defaults to nothing. but I agress we should have persistence set to `MEMORY` and `DISK`, changing

vmarkovtsev: Right, I confused them

r0mainK:
@vmarkovtsev one of the id_embeddings tests failed because we removed tensorflow, should I remove the test or put tensorflow back ?

.travis.yml
```diff
@@ -27,6 +27,8 @@ _install: &_install
   - (cd /tmp && wget "https://www.apache.org/dyn/mirrors/mirrors.cgi?action=download&filename=spark/spark-$SPARK_VERSION/$SPARK_NAME.tgz" -O $SPARK_NAME.tgz; tar -C ~/ -xvzf $SPARK_NAME.tgz)
   - export SPARK_HOME=~/$SPARK_NAME
   - export PATH=$PATH:$SPARK_HOME/bin
+  - export PYTHONPATH=$SPARK_HOME/python/:$SPARK_HOME/python/lib/py4j-0.10.4-src.zip
+  - pip install tensorflow
   - pip install -r requirements.txt codecov
```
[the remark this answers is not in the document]
r0mainK: @vmarkovtsev i did it at first, but because of tensorflow we installed `grpcio` 1.11 instead of 1.10, it did not cause problems in tests but might at some time, I thought I would move above

PR #243, opened 2018-05-09
Add Tensorflow GraphDef saving into asdf

warenlg:


sourced/ml/models/graph_def.py
```diff
@@ -0,0 +1,26 @@
+from modelforge import Model
+
+
+@register_model
+class GraphDef(Model):
+    """
+    Id split model exported a a GraphDef protobuf.
```
vmarkovtsev: The name is too generic for such a narrow model. I do not see the Tensorflow serialization here, it should be inside. Also deserialization.

So I would split it into the generic model - please do not name it GraphDef, it collides with the real class name from Tensorflow, - and the concrete model. The latter is out of this particular PRs scope so please do not add it here.

The idsplit model should be self contained. In other words, I expect it to contain the "split" method, or even `__call__`

EgorBu: I would recommend making two class: one generic to save/load tensorflow graphs, second for working with keras model as @vmarkovtsev suggested.
Second class should mimic functions from keras - just forward all arguments - it should be easy and convenient to use.
And you could create third class specific for id-splitter only to be optimized for our use cases (convert string to proper input and so on, convert predictions to proper output).

vmarkovtsev: Egor, there is no Keras here. It is the lower level GraphDef representation.

vmarkovtsev: We are using pure Tensorflow here.

warenlg:
Just to know if I'm in the right direction, is it where you want the serialization ?

sourced/ml/models/tensorflow_model.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+@register_model
+class TensorFlowModel(Model):
+    """
+    TensorFlow Protobuf model exported in the ASDS format with GraphDef inside.
+    """
+    NAME = "tensorflow-model"
+
+    def construct(self, tensorflow_model=None, session=None, list_outputs=None):
```
vmarkovtsev: Remove `tensorflow_model`. You can take it from the session.
`session: "tensorflow.Session"`
`list_outputs` > `outputs: List[str]`

Mind the strings - we do not want to rigidly depend on Tensorflow.

sourced/ml/models/tensorflow_model.py
```diff
@@ -0,0 +1,69 @@
...
+        self._tensorflow_model = tensorflow_model
+        self._session = session
+        self._list_outputs = list_outputs
+        return self
+
+    @property
+    def tensorflow_model(self):
+        """
+        Returns a tensorflow model either serialized or deserialized.
```
vmarkovtsev: There should be no "either". Always deserialized.

sourced/ml/models/tensorflow_model.py
```diff
@@ -0,0 +1,69 @@
...
+        """
+        return self._tensorflow_model
+
+    @property
+    def session(self):
```
vmarkovtsev: This is not a property of this model

sourced/ml/models/tensorflow_model.py
```diff
@@ -0,0 +1,69 @@
+from modelforge import register_model, Model
+
+
+@register_model
+class TensorFlowModel(Model):
+    """
+    TensorFlow Protobuf model exported in the ASDS format with GraphDef inside.
```
vmarkovtsev: ASDS > Modelforge

sourced/ml/models/tensorflow_model.py
```diff
@@ -0,0 +1,69 @@
...
+    """
+    NAME = "tensorflow-model"
+
+    def construct(self, tensorflow_model=None, session=None, list_outputs=None):
+        self._tensorflow_model = tensorflow_model
```
vmarkovtsev: `_tensorflow_model` > `_graphdef`

vmarkovtsev:
@warenlg I implemented this myself.

PR #246, opened 2018-05-18
Remove `**kwargs` hack and add proper filtering of arguments before call

EgorBu:
Filter arguments to a function before the call.

### DIFF
--- a/sourced/ml/transformers/basic.py
+++ b/sourced/ml/transformers/basic.py
@@ -4,6 +4,7 @@ from typing import Union
 from pyspark import RDD, Row, StorageLevel
 from pyspark.sql import DataFrame, functions
 
+from sourced.ml.extractors.helpers import filter_kwargs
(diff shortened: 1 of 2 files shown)


r0mainK:
Yes I remember this, Vadim asked me to remove it because we were switching to using `create_uast_source`, and additionnal args were filtered in functions calling `create_spark` (`create_engine` and `create_parquet_loader`). I forgot `hash` calls directly this function, so yeah I agree we should filter out extract args here too.

EgorBu:
Should we change hacks like this in [`create_engine`](https://github.com/src-d/ml/blob/master/sourced/ml/utils/engine.py#L48) & ['create_parquet_loader`](https://github.com/src-d/ml/blob/master/sourced/ml/transformers/basic.py#L230) to be consistent? Or should it be another PR for them?

vmarkovtsev:
We totally should remove them, can be here or in other PR as works better for you.

PR #250, opened 2018-05-21
Replace DataFrames with RDDs in FieldSelector and ParquetSaver

EgorBu:

### DIFF
--- a/sourced/ml/transformers/basic.py
+++ b/sourced/ml/transformers/basic.py
@@ -178,10 +178,12 @@ class FieldsSelector(Transformer):
         super().__init__(**kwargs)
         self.fields = fields
 
-    def __call__(self, df: DataFrame) -> DataFrame:
-        res = df.select(self.fields)
+    def __call__(self, rdd: RDD) -> RDD:
+        def select_fields(row):
+            return Row(**{f: getattr(row, f) for f in self.fields})
+        res = rdd.map(select_fields)
         if self.explained:
-            self._log.info("toDebugString():\n%s", res.rdd.toDebugString().decode())
+            self._log.info("toDebugString():\n%s", res.toDebugString().decode())
         return res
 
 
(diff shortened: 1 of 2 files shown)


sourced/ml/transformers/basic.py
```diff
@@ -161,10 +161,16 @@ def __init__(self, fields: Union[list, tuple], **kwargs):
...
         self.fields = fields
 
-    def __call__(self, df: DataFrame) -> DataFrame:
-        res = df.select(self.fields)
+    def __call__(self, rdd: RDD) -> RDD:
+        def select_fields(row):
+            row_dict = row.asDict()
+            res_row = {}
+            for field in self.fields:
+                res_row[field] = row_dict[field]
```
vmarkovtsev: Please use getattr and avoid copying the dict.

EgorBu:
1) Why - we are using RDDs everywhere, so these transformers became unpluggable into our pipeline.
2) Tests - definitely will be added

sourced/ml/transformers/basic.py
vmarkovtsev: @egorbu

sourced/ml/transformers/basic.py
```diff
@@ -161,10 +161,15 @@ def __init__(self, fields: Union[list, tuple], **kwargs):
         super().__init__(**kwargs)
         self.fields = fields
 
-    def __call__(self, df: DataFrame) -> DataFrame:
-        res = df.select(self.fields)
+    def __call__(self, rdd: RDD) -> RDD:
+        def select_fields(row):
+            res_row = {}
+            for field in self.fields:
+                res_row[field] = getattr(row, field)
```
vmarkovtsev: Since the tests failed and you have to change the code, how about turning this into a dict comprehension?

```python
return Row(**{f: getattr(row, f) for f in self.fields})
```

PR #251, opened 2018-05-23
Fix `Random` to `First` in DocumentFrequencies.dump()

zurk:
I noticed that we have First 10 identifiers output in code, but the title says it is random.
So I have two ways to solve it:
1. Make random in both places.
2. Make first in both places.

I decide to choose the 2nd option. I think it is better and easier to compare models or track model changes after modifications.


vmarkovtsev:
@zurk It prints first 10 identifiers from `dict.items()`. The order is different every time.

PR #252, opened 2018-05-23
Add subcommand to merge Document Frequencies model

zurk:
Since we need to merge the models, as a spark is not reliable enough I add `merge_df` subcommand.
As was discussed I take the old code for Document Frequencies model because it is small enough to be processed by one node. 

I make some minor modifications just to integrate old code to the new codebase and write more tests in `test_merge_df.py`.

The biggest files `model_converters/base.py` and `test_model2base.py` are mostly the same.

### DIFF
--- a/sourced/ml/__main__.py
+++ b/sourced/ml/__main__.py
@@ -9,9 +9,10 @@ from sourced.ml.extractors import IdentifierDistance
 from sourced.ml.cmd import bigartm2asdf_entry, dump_model, projector_entry, bow2vw_entry, \
     run_swivel, postprocess_id2vec, preprocess_id2vec, repos2coocc_entry, repos2df_entry, \
     repos2ids_entry, repos2bow_entry, repos2roles_and_ids_entry, repos2id_distance_entry, \
-    repos2id_sequence_entry
+    repos2id_sequence_entry, merge_df_entry
 from sourced.ml.cmd.args import add_df_args, add_feature_args, add_split_stem_arg, \
-    add_repo2_args, add_bow_args, add_repartitioner_arg, ArgumentDefaultsHelpFormatterNoNone
+    add_vocabulary_size_arg, add_repo2_args, add_bow_args, add_repartitioner_arg, add_filter_arg, \
+    add_min_docfreq, ArgumentDefaultsHelpFormatterNoNone
 from sourced.ml.cmd.run_swivel import mirror_tf_args
 from sourced.ml.utils import install_bigartm
 
@@ -186,6 +187,22 @@ def get_parser() -> argparse.ArgumentParser:
         "input", help="Path to the model file, URL or UUID.")
     dump_parser.add_argument("--gcs", default=None, dest="gcs_bucket",
                              help="GCS bucket to use.")
+    # ------------------------------------------------------------------------
+    merge_df = add_parser("merge-df", "Merge DocumentFrequency models to a singe one.")
+    merge_df.set_defaults(handler=merge_df_entry)
+    add_filter_arg(merge_df)
+    add_min_docfreq(merge_df)
+    add_vocabulary_size_arg(merge_df)
+    merge_df.add_argument(
+        "-o", "--output", required=True,
+        help="Path to the merged document frequencies model.")
+    merge_df.add_argument(
+        "-i", "--input", required=True,
+        help="Directory to scan recursively for asdf files.")
+    merge_df.add_argument(
+        "--ordered", action="store_true", default=False,
+        help="Save OrderedDocumentFrequencies. "
+             "If not specified DocumentFrequency model will be saved")
 
     return parser
 
--- a/sourced/ml/__main__.py
+++ b/sourced/ml/__main__.py
@@ -188,21 +188,22 @@ def get_parser() -> argparse.ArgumentParser:
     dump_parser.add_argument("--gcs", default=None, dest="gcs_bucket",
                              help="GCS bucket to use.")
     # ------------------------------------------------------------------------
-    merge_df = add_parser("merge-df", "Merge DocumentFrequency models to a singe one.")
+    merge_df = add_parser("merge-df", "Merge DocumentFrequencies models to a singe one.")
     merge_df.set_defaults(handler=merge_df_entry)
     add_filter_arg(merge_df)
     add_min_docfreq(merge_df)
     add_vocabulary_size_arg(merge_df)
     merge_df.add_argument(
         "-o", "--output", required=True,
-        help="Path to the merged document frequencies model.")
+        help="Path to the merged DocumentFrequencies model.")
     merge_df.add_argument(
-        "-i", "--input", required=True,
-        help="Directory to scan recursively for asdf files.")
+        "-i", "--input", required=True, nargs="+",
+        help="DocumentFrequencies models input files."
+             "Use `-i -` to read input files from stdin.")
     merge_df.add_argument(
         "--ordered", action="store_true", default=False,
         help="Save OrderedDocumentFrequencies. "
-             "If not specified DocumentFrequency model will be saved")
+             "If not specified DocumentFrequencies model will be saved")
 
     return parser
 
--- a/sourced/ml/cmd/args.py
+++ b/sourced/ml/cmd/args.py
@@ -1,9 +1,17 @@
 import argparse
 import json
+import logging
+from pathlib import Path
+from typing import Optional
+import sys
+
 
 from sourced.ml import extractors
 from sourced.ml.transformers import BOWWriter, Moder
 from sourced.ml.utils import add_engine_args
+
+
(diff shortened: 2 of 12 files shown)


sourced/ml/__main__.py
```diff
@@ -192,6 +192,22 @@ def add_parser(name, help_message):
         "input", help="Path to the model file, URL or UUID.")
     dump_parser.add_argument("--gcs", default=None, dest="gcs_bucket",
                              help="GCS bucket to use.")
+    # ------------------------------------------------------------------------
+    merge_df = add_parser("merge_df", "Merge DocumentFrequency models to a singe one.")
```
vmarkovtsev: into a single one

vmarkovtsev: How about `mdf` instead of `merge_df`? Or `merge-df`

sourced/ml/__main__.py
```diff
@@ -192,6 +192,22 @@ def add_parser(name, help_message):
...
+    merge_df.set_defaults(handler=merge_df_entry)
+    add_filter_arg(merge_df)
+    add_min_docfreq(merge_df)
+    add_vocabulary_size_arg(merge_df)
+    merge_df.add_argument(
+        "-o", "--output", required=True,
+        help="Path to the merged document frequencies model.")
+    merge_df.add_argument(
+        "-i", "--input", required=True,
+        help="Directory to scan recursively for asdf files.")
```
vmarkovtsev: `asdf` > `DocumentFrequencies`

sourced/ml/__main__.py
```diff
@@ -192,6 +192,22 @@ def add_parser(name, help_message):
...
+    merge_df.add_argument(
+        "--ordered", action="store_true", default=False,
+        help="Save OrderedDocumentFrequency. "
+             "If not specified DocumentFrequency model will be saved")
```
vmarkovtsev: Frequenc**ies**

sourced/ml/__main__.py
```diff
@@ -192,6 +192,22 @@ def add_parser(name, help_message):
...
+    # ------------------------------------------------------------------------
+    merge_df = add_parser("merge_df", "Merge DocumentFrequency models to a singe one.")
+    merge_df.set_defaults(handler=merge_df_entry)
+    add_filter_arg(merge_df)
+    add_min_docfreq(merge_df)
+    add_vocabulary_size_arg(merge_df)
+    merge_df.add_argument(
+        "-o", "--output", required=True,
+        help="Path to the merged document frequencies model.")
+    merge_df.add_argument(
```
vmarkovtsev: How about taking a list of files as an input? Recursive scan can easily be performed with `find` and the whole thing can be run as

```
find '*.asdf' | srcml mdf -i - -o result.asdf
```

Otherwise it is impossible to merge only a few specific files, only the directory.

sourced/ml/cmd_entries/merge_df.py
```diff
@@ -0,0 +1,8 @@
+from sourced.ml.models import MergeDocFreq
+
+
+def merge_df_entry(args):
+    merger = MergeDocFreq(ordered=args.ordered,
+                          vocabulary_size=args.vocabulary_size,
+                          min_docfreq=args.min_docfreq)
+    merger.convert(args.input, args.output, pattern=args.filter)
```
vmarkovtsev: This class will be simplified too - you will feed only the list of input files and the output name. The compilation of that list is outside of the class.

sourced/ml/models/model_converters/base.py
```diff
@@ -0,0 +1,128 @@
...
+from pathlib import Path
+from typing import Union
+
+from modelforge import Model
+from modelforge.progress_bar import progress_bar
+
+from sourced.ml.utils.pickleable_logger import PickleableLogger
+
+
+class Model2Base(PickleableLogger):
```
vmarkovtsev: Was it copied from the old codebase? What has been changed?

sourced/ml/models/model_converters/merge_df.py
```diff
@@ -0,0 +1,45 @@
...
+        df_model = OrderedDocumentFrequencies if self.ordered else DocumentFrequencies
+        df_model(log_level=self._log.level) \
+            .construct(self._docs, self._df) \
+            .prune(self.min_docfreq) \
+            .greatest(self.vocabulary_size) \
+            .save(self._save_path(index, destdir))
+
+    @staticmethod
+    def _save_path(index: int, destdir: str):
+        if destdir.endswith(".asdf"):
```
vmarkovtsev: I am not happy that we hardcode asdf here and in the other places. It should be an implementation detail of Modelforge.

I suggest to create `modelforge.Model.FILE_EXTENSION` and use it everywhere.

sourced/ml/__main__.py
zurk: ok, 'merge-df'

sourced/ml/__main__.py
zurk: Actually, you can do it with `--filter` flag.
Maybe it is better to keep both options? 
you can specify `-i -` to read a list of files from stdin or provide a directory `-i dirpath`

sourced/ml/models/model_converters/base.py
zurk: Yes, I mentioned it in the PR description. 
Maybe I change few lines to make it work, but nothing more.

sourced/ml/models/model_converters/merge_df.py
zurk: good point! I prefer to make an another to PRs to modelforge and ml for this change.

sourced/ml/__main__.py
vmarkovtsev: This is called the "UNIX way" :) You can remove all that code and do not support it with tests.
Anyway, the stdin feature should be here.

vmarkovtsev:
@zurk Conflicts

zurk:
Solved

sourced/ml/__main__.py
```diff
@@ -186,6 +187,22 @@ def add_parser(name, help_message):
         "input", help="Path to the model file, URL or UUID.")
     dump_parser.add_argument("--gcs", default=None, dest="gcs_bucket",
                              help="GCS bucket to use.")
+    # ------------------------------------------------------------------------
+    merge_df = add_parser("merge-df", "Merge DocumentFrequency models to a singe one.")
```
vmarkovtsev: DocumentFrequencies

sourced/ml/__main__.py
```diff
@@ -186,6 +187,22 @@ def add_parser(name, help_message):
...
+    merge_df.set_defaults(handler=merge_df_entry)
+    add_filter_arg(merge_df)
+    add_min_docfreq(merge_df)
+    add_vocabulary_size_arg(merge_df)
+    merge_df.add_argument(
+        "-o", "--output", required=True,
+        help="Path to the merged document frequencies model.")
```
vmarkovtsev: DocumentFrequencies

We should use exactly the same class name everywhere to reduce confusion

sourced/ml/__main__.py
vmarkovtsev: @zurk not addressed

sourced/ml/cmd/args.py
```diff
@@ -17,6 +25,30 @@ def _get_help_string(self, action):
...
+    :return: an iterator over input files.
+    """
+    log = log.info if log else (lambda *x: None)
+    if input_arg == "-":
+        log("Reading file paths from stdin.")
+        for line in sys.stdin:
+            yield line.strip()
+    else:
+        log("Scanning %s", input_arg)
+        if filter_arg:
```
vmarkovtsev: If there is no filter arg we should `return input_arg` (provided by `nargs="+"`)

sourced/ml/cmd/args.py
```diff
@@ -17,6 +25,30 @@ def _get_help_string(self, action):
...
+            for path in Path(input_arg).glob(filter_arg):
```
vmarkovtsev: That is,

```python
for arg in input_arg:
    if path.isfile(arg):
        yield arg
    elif filter_arg:
        for path in Path(arg).glob(filter_arg):
            yield str(path)
```

zurk: I want to process directories without filtering also, so I change it to:
```python
        for arg in input_arg:
            if os.path.isfile(arg):
                yield arg
            else:
                if filter_arg:
                    for path in Path(arg).glob(filter_arg):
                        yield str(path)
                elif os.path.isdir(arg):
                    for path in os.listdir(arg):
                        yield path
```

vmarkovtsev: See how complex it became?

This is why we need to nuke all this code and leave

```python
for arg in input_arg:
    yield arg
```

The rest is GNU find's job. I override you here, sorry, but it really must be done this way.

PR #254, opened 2018-05-24
Add coocc model merger

zurk:
Depends on https://github.com/src-d/ml/pull/252, but they are not tightly coupled.

Since we need to merge the models, as a spark is not reliable enough I add `merge_coocc` subcommand.

As was discussed I use spark to process data because it can be big enough not to be processed by one node. I decided not to add another Transformer because this subcommand is quite different from our `repo2` commands.

### DIFF
--- a/sourced/ml/cmd/merge_coocc.py
+++ b/sourced/ml/cmd/merge_coocc.py
@@ -1,0 +1,75 @@
+import logging
+import operator
+from uuid import uuid4
+
+from modelforge.progress_bar import progress_bar
+import numpy as np
+from scipy.sparse import coo_matrix
+
+from sourced.ml.cmd.args import handle_input_arg
+from sourced.ml.extractors.helpers import filter_kwargs
+from sourced.ml.models import OrderedDocumentFrequencies, Cooccurrences
+from sourced.ml.transformers import CooccModelSaver
+from sourced.ml.utils.engine import pause
+from sourced.ml.utils.spark import create_spark
+
+
+MAX_INT32 = 2**31 - 1
+
+
+@pause
+def merge_coocc_entry(args):
+    log = logging.getLogger("merge_coocc")
+    log.setLevel(args.log_level)
+    filepaths = list(handle_input_arg(args.input, log))
+    log.info("Found %d files", len(filepaths))
+    df = OrderedDocumentFrequencies().load(args.docfreq)
+    if args.no_spark:
+        merge_coocc_entry_no_spark(df, filepaths, log, args)
+    else:
+        merge_coocc_entry_spark(df, filepaths, log, args)
+
+
+def merge_coocc_entry_spark(df, filepaths, log, args):
+    session_name = "merge_coocc-%s" % uuid4()
+    session = create_spark(session_name, **filter_kwargs(args.__dict__, create_spark))
+    spark_context = session.sparkContext
+    global_index = spark_context.broadcast(df.order)
+
+    coocc_rdds = []
+    for path in progress_bar(filepaths, log):
+        coocc = Cooccurrences().load(path)
+        rdd = coocc.matrix_to_rdd(spark_context)  # rdd structure: ((row, col), weight)
+        tokens = spark_context.broadcast(coocc.tokens)
+        coocc_rdds.append(
+            rdd.map(lambda row: ((global_index.value.get(tokens.value[row[0][0]], -1),
+                                  global_index.value.get(tokens.value[row[0][1]], -1)),
+                                 np.uint32(row[1])))
+               .filter(lambda row: row[0][0] >= 0))
+
+    log.info("Union of concurrence matrices...")
+    rdd = spark_context \
+        .union(coocc_rdds) \
+        .reduceByKey(lambda x, y: min(MAX_INT32, x + y))
+    CooccModelSaver(args.output, df)(rdd)
+
+
+def merge_coocc_entry_no_spark(df, filepaths, log, args):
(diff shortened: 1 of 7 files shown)


vmarkovtsev:
@zurk please rebase

zurk:
Rebased

vmarkovtsev:
@zurk merge conflict here

sourced/ml/cmd/merge_coocc.py
```diff
@@ -0,0 +1,75 @@
...
+    global_index = spark_context.broadcast(df.order)
+
+    coocc_rdds = []
+    for path in progress_bar(filepaths, log):
+        coocc = Cooccurrences().load(path)
+        rdd = coocc.matrix_to_rdd(spark_context)  # rdd structure: ((row, col), weight)
+        tokens = spark_context.broadcast(coocc.tokens)
+        coocc_rdds.append(
+            rdd.map(lambda row: ((global_index.value.get(tokens.value[row[0][0]], -1),
+                                  global_index.value.get(tokens.value[row[0][1]], -1)),
```
vmarkovtsev: `tuple(global_index.value.get(tokens.value[row[0][i]], -1) for i in (0, 1))`

vmarkovtsev: It would also help to have a comment here which explains this monster's structure or gives an example

sourced/ml/cmd/merge_coocc.py
```diff
@@ -0,0 +1,75 @@
...
+                                 np.uint32(row[1])))
+               .filter(lambda row: row[0][0] >= 0))
+
+    log.info("Union of concurrence matrices...")
```
vmarkovtsev: Calculating the union of **cooccurrence** matrices

sourced/ml/cmd/merge_coocc.py
```diff
@@ -0,0 +1,75 @@
...
+    rdd = spark_context \
+        .union(coocc_rdds) \
+        .reduceByKey(lambda x, y: min(MAX_INT32, x + y))
+    CooccModelSaver(args.output, df)(rdd)
+
+
+def merge_coocc_entry_no_spark(df, filepaths, log, args):
+    log.info("Without spark")
```
vmarkovtsev: "Merging cooccurrences without using PySpark"

sourced/ml/cmd/merge_coocc.py
```diff
@@ -0,0 +1,75 @@
...
+    shape = (len(df) + 1, len(df) + 1)
```
vmarkovtsev: `(len(df) + 1,) * 2`

sourced/ml/cmd/merge_coocc.py
```diff
@@ -0,0 +1,75 @@
...
+        coocc = Cooccurrences().load(path)
+        index = [df.order.get(x, len(df) + 1) for x in coocc.tokens]
+        rows = [index[x] for x in coocc.matrix.row]
+        cols = [index[x] for x in coocc.matrix.col]
+        result += coo_matrix(
+            (coocc.matrix.data, (rows, cols)), shape=shape, dtype=np.uint32)
+        indx_overflow = np.where(result.data > MAX_INT32)
+        if indx_overflow[0].size > 0:
+            log.warning("Overflow in %d elements."
+                        "They will be saturated to %d" % (indx_overflow[0].size, MAX_INT32))
```
vmarkovtsev: "Saturating them to MAX_INT32." We do not really need to read these 2 billions :)

sourced/ml/cmd/merge_coocc.py
```diff
@@ -0,0 +1,75 @@
...
+        .union(coocc_rdds) \
+        .reduceByKey(lambda x, y: min(MAX_INT32, x + y))
+    CooccModelSaver(args.output, df)(rdd)
+
+
+def merge_coocc_entry_no_spark(df, filepaths, log, args):
+    log.info("Without spark")
+    shape = (len(df) + 1, len(df) + 1)
+    result = coo_matrix(shape, dtype=np.uint32)
+    for path in progress_bar(filepaths, log):
```
vmarkovtsev: You should put a comment here which explains the algorithm.

1. Although we store uint32, we actually never have elements greater than MAX_INT32
2. We assume that both `result` and the summed matrix do not have elements greater than MAX_INT32 (**btw, you do not check it in your code for the summed matrix**)
3. As soon as we have a value bigger than MAX_INT32 after summing, we saturate
4. Thus we lose 2x data range but do not allocate any memory and it works faster than MAX_UINT32 checks
5. Only x (**substitute x with experimental data**) elements saturate in PGA so this is fine

sourced/ml/tests/test_merge_coocc_entry.py
```diff
@@ -0,0 +1,64 @@
...
+        output = os.path.join(_input_dir, "res_coocc.asdf")
+        docfreq = models.COOCC_DF
+        pause = False
+        filter = "**/*.asdf"
+        log_level = "INFO"
+        no_spark = _no_spark
+    return args
+
+
+class MergeCooccEntry(unittest.TestCase):
```
vmarkovtsev: Is there a test which checks the saturation?

zurk:
Rebased

sourced/ml/tests/test_merge_coocc_entry.py
zurk: No, will add

vmarkovtsev:
Congrats!

PR #255, opened 2018-05-24
[id-split] part 1: feature extraction

EgorBu:
* feature extraction
* tests
* misc for travis and setup

PR #256, opened 2018-05-25
[id-split] part 2: CNN/RNN models

EgorBu:
Add functionality to create CNN/RNN models
* CNN builder
* RNN builder
* metrics
* tests for metrics

Should be merged after https://github.com/src-d/ml/pull/255

### DIFF
--- a/sourced/ml/algorithms/id_splitter/nn_model.py
+++ b/sourced/ml/algorithms/id_splitter/nn_model.py
@@ -1,0 +1,211 @@
+import warnings
+
+import numpy as np
+try:
+    import keras
+    from keras import backend as K
+    from keras.layers import BatchNormalization, Concatenate, Conv1D, CuDNNGRU, CuDNNLSTM, Dense, \
+        Embedding, GRU, Input, LSTM, TimeDistributed
+    from keras.models import Model
+    import tensorflow as tf
+except ImportError as e:
+    warnings.warn("Tensorflow or/and Keras are not installed, dependent functionality is "
+                  "unavailable.")
+
+
+LOSS = "binary_crossentropy"
+METRICS = ["accuracy"]
+DEFAULT_RNN_TYPE = "LSTM"
+
+
+def register_metric(metric):
+    assert isinstance(metric, str) or callable(metric)
+    METRICS.append(metric)
+    return metric
+
+
+def prepare_input_emb(maxlen, n_uniq):
+    """
+    One-hot encoding of characters
+    :param maxlen: maximum length of input sequence
+    :param n_uniq: number of unique characters
+    :return: tensor for input, one-hot character embeddings
+    """
+    char_seq = Input((maxlen,))
+    emb = Embedding(input_dim=n_uniq + 1, output_dim=n_uniq + 1, input_length=maxlen,
+                    mask_zero=False, weights=[np.eye(n_uniq + 1)], trainable=False)(char_seq)
+    return char_seq, emb
+
+
+def add_output_layer(input_layer):
+    """
+    Output layer has 1 sigmoid per character that should predict if there's a space before char
+    :param input_layer: hidden layer before output layer
+    :return: layer
+    """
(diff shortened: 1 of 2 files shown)


EgorBu:
Coverage should be increased in next PR

sourced/ml/algorithms/id_splitter/nn_model.py
```diff
@@ -0,0 +1,211 @@
...
+def precision(y_true, y_pred):
+    """
+    Precision metric. Only computes a batch-wise average of recall.
+    :param y_true: tensor
+    :param y_pred: tensor
+    :return: tensor
+    """
+    true_positives = K.sum(K.round(K.clip(y_true * y_pred, 0, 1)))
+    predicted_positives = K.sum(K.round(K.clip(y_pred, 0, 1)))
+    precision = true_positives / (predicted_positives + K.epsilon())
```
r0mainK: no need to create the precision variable, might as well directly return it

sourced/ml/algorithms/id_splitter/nn_model.py
```diff
@@ -0,0 +1,211 @@
...
+def recall(y_true, y_pred):
+    """
+    Recall metric. Only computes a batch-wise average of recall.
+    :param y_true: tensor
+    :param y_pred: tensor
+    :return: tensor
+    """
+    true_positives = K.sum(K.round(K.clip(y_true * y_pred, 0, 1)))
+    possible_positives = K.sum(K.round(K.clip(y_true, 0, 1)))
+    recall = true_positives / (possible_positives + K.epsilon())
```
r0mainK: same here, no need to create recall variable, also `possible_positives` could be refactored to `positives`, or `actual_positives` ? i find it kinda misguiding at first glance, as possible kinda infers lack of certainty, which there is not in this case since we know they are positives

EgorBu: https://en.wikipedia.org/wiki/Precision_and_recall
`possible_positives` - it could be `relevant_positive` / `retrieved_positive` if we want to follow wiki naming.
I don't mind renaming - let's see which option is more preferable by other reviewers.

sourced/ml/algorithms/id_splitter/nn_model.py
```diff
@@ -0,0 +1,233 @@
+import warnings
+
+import numpy
+import keras
+from keras import backend as K
```
vmarkovtsev: We should not do this import in production code. Change to `kbackend`

sourced/ml/algorithms/id_splitter/nn_model.py
```diff
@@ -0,0 +1,233 @@
...
+    :param X: input layer.
+    :param units: number of neurons in the output layer.
+    :param rnn_layer: type of cell in the RNN.
+    :param dev0: device that will be used for forward pass of RNN and concatenation.
+    :param dev1: device that will be used for backward pass.
+
+    :return: output bidirectional RNN layer.
+    """
+    # select the RNN layer
+    rnn_layer_mapping = {"GRU": GRU, "LSTM": LSTM, "CuDNNLSTM": CuDNNLSTM, "CuDNNGRU": CuDNNGRU}
```
vmarkovtsev: getattr

sourced/ml/algorithms/id_splitter/nn_model.py
```diff
@@ -0,0 +1,233 @@
...
+    """
+    Builds a single convolutional layer.
+
+    :param X: input layer.
+    :param filters: number of output filters in the convolution.
+    :param kernel_sizes: list of lengths of the 1D convolution window.
+    :param output_n_filters: number of 1D output filters.
+    :return: output layer.
+    """
+    # normalize of the input
```
vmarkovtsev: ~of~

vmarkovtsev:
@warenlg Waiting for the fix of Travis.

PR #258, opened 2018-05-28
Add handling of read errors to spark conf

r0mainK:
We raised [this issue](https://github.com/src-d/engine/issues/385) which was addressed in [this PR](https://github.com/src-d/engine/pull/395) which just got merged, this PR just ports the change to the Spark configuration in ml.


fulaphex:
Why do we prefer to set this option in SparkSession builder instead of setting them while initializing the `Engine` object? Same for `skip_cleanup`

PR #260, opened 2018-05-30
Add create_or_load_ordered_df function

r0mainK:
**CONTEXT**

While running apollo on PGA, my job failed after ~36h, out of which 20h were spent creating then writing the docfreq model. The reason for failure seems to be due to Spark instability with ml when processing very large amounts of data (shuffling of ~1TB happened multipl times ...). So I relaunched it, but instead of recalculating the ordered docfreq, I simply edited the code to not waste 20h recreating the already existing ordered docfreq model. I think that this should be integrated to the code, as in my case it halfs processing time, which is rather long - and from what I saw it can be used for both repos2cooc and repos2bow. 

**CHANGES**

- I added the file `utils/docfreq` for the `create_or_load_ordered_df` function used both in repos2cooc and repos2bow. It takes as input the `args`, logger of calling function, the transformer containing bag features and the number of docs. The function either loads an existing model, or creates and saves one from the transformer.
- I updated the `add_df` function in `cmd_entries/args`, creating an exclusive group to have either the `--docfreq-in` flag if loading or `--docfreq-out` if creating/saving. I thought it would be more explicit then checking the filepath.
- I updated the main, parsers for `preprocess_id2vec` and `projector_entry` also had `--docfreq` flag, thought it would be logical to make use of the parser, even though some args will of the namespace will not be used
- I integrated new function in  `repos2cooc` and`repos2bow`, and change the names of the args in `preprocess_id2vec`, `projector_entry` and `repos2df`
- Added a test for new function

### DIFF
--- a/sourced/ml/utils/docfreq.py
+++ b/sourced/ml/utils/docfreq.py
@@ -1,0 +1,34 @@
+import logging
+
+from sourced.ml.transformers import BagFeatures2DocFreq, Uast2BagFeatures
+from sourced.ml.models import OrderedDocumentFrequencies
+
+
+def create_or_load_ordered_df(args, ndocs: int, bag_features: Uast2BagFeatures):
+    """
+    Returns a preexisting OrderedDocumentFrequencies model from docfreq_in, or generates one
+    from the flattened bags of features using args and saves it to docfreq_out.
+
+    :param args: Instance of `argparse.Namespace` that contains docfreq_in, docfreq_out,
+                 min_docfreq, and vocabulary_size.
+    :param ndocs: Number of documents (can be repos, files or functions)
+    :param bag_features: Transformer containing bags of features extracted from the data (the call
+                         instantiates an RDD: [(key, doc), val] where key is a specific feature
+                         that appeared val times in the document doc.
+    :return: OrderedDocumentFrequencies model
+    """
+    log = logging.getLogger("create_or_load_ordered_df")
+    if args.docfreq_in:
+        log.info("Loading ordered docfreq model from %s ...", args.docfreq_in)
+        return OrderedDocumentFrequencies().load(args.docfreq_in)
+    log.info("Calculating the document frequencies, hold tight ...")
+    df = bag_features \
+        .link(BagFeatures2DocFreq()) \
+        .execute()
+    log.info("Writing ordered docfreq model to %s ...", args.docfreq_out)
+    df_model = OrderedDocumentFrequencies() \
+        .construct(ndocs, df) \
+        .prune(args.min_docfreq) \
+        .greatest(args.vocabulary_size) \
+        .save(args.docfreq_out)
+    return df_model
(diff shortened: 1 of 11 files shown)


sourced/ml/cmd_entries/args.py
```diff
@@ -57,13 +57,15 @@ def add_repo2_args(my_parser: argparse.ArgumentParser, default_packages=None):
...
+def add_df_args(my_parser: argparse.ArgumentParser, required=True):
     my_parser.add_argument(
         "--min-docfreq", default=1, type=int,
         help="The minimum document frequency of each feature.")
-    my_parser.add_argument(
-        "--docfreq", required=True,
-        help="[OUT] The path to the OrderedDocumentFrequencies model.")
+    df_group = my_parser.add_mutually_exclusive_group(required=required)
+    df_group.add_argument(
+        "--docfreq-out", help="Path to save docfreq model.")
```
vmarkovtsev: better "DocumentFrequencies" instead of "docfreq" in `help`

sourced/ml/utils/docfreq.py
```diff
@@ -0,0 +1,21 @@
+from sourced.ml.transformers import BagFeatures2DocFreq
+from sourced.ml.models import OrderedDocumentFrequencies
+
+
+def create_or_load_ordered_df(args, ndocs, uast_extractor, log):
+    if args.docfreq_in:
+        log.info("Loading ordered docfreq model from %s ...", args.docfreq_in)
+        return OrderedDocumentFrequencies() \
```
vmarkovtsev: You don't really need "\\" here

sourced/ml/utils/docfreq.py
```diff
@@ -0,0 +1,21 @@
...
+            .load(args.docfreq_in)
+    else:
```
vmarkovtsev: `else` can be removed - you make a `return` in the other branch

sourced/ml/utils/docfreq.py
```diff
@@ -0,0 +1,21 @@
+from sourced.ml.transformers import BagFeatures2DocFreq
+from sourced.ml.models import OrderedDocumentFrequencies
+
+
+def create_or_load_ordered_df(args, ndocs, uast_extractor, log):
```
vmarkovtsev: Would be nice to have a comment what this function does

sourced/ml/cmd_entries/args.py
vmarkovtsev: Also we need to mention that we actually generate it which can take some time (as you described)

r0mainK: done

r0mainK:
Rebased off https://github.com/src-d/ml/pull/264 since it contains fix for broken tests, should be merged after that one (and a rebase)

vmarkovtsev:
@r0mainK Can you please rebase one more time

r0mainK:
@vmarkovtsev done

PR #262, opened 2018-06-06
Add co-occurrence matrix values check to id2vec_preproc

zurk:
Raise an error if matrix contains negative elements


sourced/ml/cmd/preprocess_id2vec.py
```diff
@@ -37,6 +37,9 @@ def preprocess_id2vec(args):
     log.info("Loading docfreq model from %s", args.docfreq_in)
     df_model = DocumentFrequencies(log_level=args.log_level).load(source=args.docfreq_in)
     coocc_model = Cooccurrences().load(args.input)
+    if numpy.any(coocc_model.matrix.data < 0):
+        raise ValueError(("Co-occurrence matrix %s contains negative elements. "
+                          "Please check its correctness.") % args.input)
```
vmarkovtsev: Sorry for being a pedantic pain in the ass.

its -> it's

Besides, it makes sense to check against NaN as well.

PR #267, opened 2018-06-13
Move travis spark version to 2.2.1

zurk:
Since spark v2.2.0 disappears from http://apache-mirror.rbc.ru/pub/apache/spark/ Travis failed and cannot download it anymore and Travis CI failes. Also, I asked DR team about spark version for engine. They said it should be v2.2.1. It will be fixed soon for python package (https://github.com/src-d/engine/pull/401) also.

We can not just exclude spark from travis because a lot of tests use it.

After this PR is merged Travis should work again.


setup.py
```diff
@@ -31,7 +31,6 @@
                       "bblfsh>=2.2.1,<3.0",
                       "modelforge>=0.5.4-alpha",
                       "sourced-engine>=0.5.1,<0.7",
-                      "pyspark>=2.2.0,<2.2.1",
```
vmarkovtsev: Should be `>=2.2.1,<2.3` if works

.travis.yml
```diff
@@ -16,7 +16,7 @@ _install: &_install
   - source ~/.gimme/envs/latest.env
   - pip install --upgrade pip
   - export HADOOP_VERSION=2.7
-  - export SPARK_VERSION=2.2.0
+  - export SPARK_VERSION=2.2.1
   - export SPARK_NAME=spark-$SPARK_VERSION-bin-hadoop$HADOOP_VERSION
   - (cd /tmp && wget "https://www.apache.org/dyn/mirrors/mirrors.cgi?action=download&filename=spark/spark-$SPARK_VERSION/$SPARK_NAME.tgz" -O $SPARK_NAME.tgz; tar -C ~/ -xvzf $SPARK_NAME.tgz)
```
vmarkovtsev: As discussed, try to remove this and see if the tests pass

setup.py
zurk: but we have pyspark requirements from engine https://github.com/src-d/engine/blob/master/python/setup.py#L29 
should we fix anything in our setup?

vmarkovtsev: Right. I thought we removed this a long time, weird

PR #269, opened 2018-06-14
Cmd module refactoring

zurk:
The main reason for this PR is a regular change of lines in `__main__` imports:
https://github.com/src-d/ml/blob/master/sourced/ml/__main__.py#L10-L17
We update it frequently, we have more and more lines here (8 already! ) so it is better to use 
```python
from sourced.ml import cmd
from sourced.ml.cmd import args
```
instead of 
```python
from sourced.ml.cmd import ...
from sourced.ml.cmd.args import ...
```

I considered to use `from sourced.ml.cmd import *` but it is not so good way for many reasons. 

So as a result:
1. I change import of `cmd` and `args` modules in `__main__.py`
2. because of that we do not need `_entry` prefix for `cmd` functions, now we call them as `cmd.function`, so I removed it.
3. I notice that we have inconsistency for id2vec files and function names with cmd subcommands. So I change `_id2vec` postfix to `id2vec_` prefix in cmd functions and files.

Comments are welcome!


sourced/ml/__main__.py
```diff
@@ -53,26 +47,26 @@ def add_parser(name, help_message):
...
     repos2bow_parser = add_parser(
         "repos2bow", "Convert source code to the bag-of-words model.")
-    repos2bow_parser.set_defaults(handler=repos2bow_entry)
-    add_df_args(repos2bow_parser)
-    add_repo2_args(repos2bow_parser)
-    add_feature_args(repos2bow_parser)
-    add_bow_args(repos2bow_parser)
-    add_repartitioner_arg(repos2bow_parser)
+    repos2bow_parser.set_defaults(handler=cmd.repos2bow)
+    args.add_df_args(repos2bow_parser)
```
vmarkovtsev: Suggestion for the future:

```python
for add in (args.add_df_args, args.add_repo2_args, args.add_feature_args, args.add_bow_args, ...):
    add(my_parser)
```

sourced/ml/cmd/id2vec_projector.py
```diff
@@ -5,7 +5,7 @@
 from sourced.ml.models import Id2Vec
 
 
-def projector_entry(args):
+def id2vec_projector(args):
```
vmarkovtsev: Names should be verb-like: `id2vec_project`

vmarkovtsev:
@zurk Merging with a few minor suggestions which can be addressed later.

PR #275, opened 2018-06-18
Add option to pickle doc index and stop_pipeline of repo2bow

r0mainK:
[Linked PR in apollo](https://github.com/src-d/apollo/pull/61/)

**IDEA**

So since on huge batches this has not proven robust for now, and given the point of failure is during the calculation of TermFrequency right before TFIDF, I think we can run once on all the data to get the DocFreq model as well as a pickle of the index of all documents.
With these available, we can run on subsets of the total without loosing any  - the only thing is not to forget to rename the output bags or they will be overwritten.

**CONTEXT**

Don't know if this works although it should, running it tonight/tomorrow. Would like input if possible.

**CHANGES**

- main: added pickle_index as an optional input or output path. if supplied the pickle will be loaded/saved to the location
- transformers/index: logic of above functionality
- repos2bow: initialized the document indexer with the pickle feature, added an arg to the template to skips bags and metadata writing, focusing on the document_indexer and docfreq model.

EDIT: tomorrow will probably add stuff

### DIFF
--- a/sourced/ml/cmd/repos2bow.py
+++ b/sourced/ml/cmd/repos2bow.py
@@ -2,12 +2,12 @@ import logging
 from uuid import uuid4
 
 from sourced.ml.extractors import create_extractors_from_args
-from sourced.ml.models import QuantizationLevels
-from sourced.ml.transformers import UastDeserializer, Uast2Quant, BagFeatures2TermFreq, \
-    Uast2BagFeatures, HeadFiles, TFIDF, Cacher, Indexer, UastRow2Document, BOWWriter, Moder, \
-    create_uast_source, Repartitioner
+from sourced.ml.transformers import UastDeserializer, BagFeatures2TermFreq, Uast2BagFeatures, \
+    HeadFiles, TFIDF, Cacher, Indexer, UastRow2Document, BOWWriter, Moder, create_uast_source, \
+    Repartitioner
 from sourced.ml.utils.engine import pipeline_graph, pause
 from sourced.ml.utils.docfreq import create_or_load_ordered_df
+from sourced.ml.utils.quant import create_or_apply_quant
 
 
 @pause
@@ -25,21 +25,18 @@ def repos2bow_template(args, select=HeadFiles, cache_hook=None, save_hook=None):
     # Row items.
     uast_extractor = uast_extractor.link(UastRow2Document())
     log.info("Extracting UASTs and indexing documents...")
-    document_indexer = Indexer(Uast2BagFeatures.Columns.document)
-    uast_extractor.link(document_indexer).execute()
+    document_indexer = Indexer(Uast2BagFeatures.Columns.document,
+                               cached_index_path=args.cached_index_path)
+    if args.index_in is None:
+        uast_extractor.link(document_indexer).execute()
     ndocs = len(document_indexer)
     log.info("Number of documents: %d", ndocs)
     uast_extractor = uast_extractor.link(UastDeserializer())
-    quant = Uast2Quant(extractors)
-    uast_extractor.link(quant).execute()
-    if quant.levels:
-        log.info("Writing quantization levels to %s", args.quant)
-        QuantizationLevels().construct(quant.levels).save(args.quant)
+    if args.quant:
+        create_or_apply_quant(args.quant, extractors, uast_extractor)
     uast_extractor = uast_extractor \
         .link(Uast2BagFeatures(extractors))
-
     df_model = create_or_load_ordered_df(args, ndocs, uast_extractor)
-
     bags_writer = uast_extractor \
         .link(BagFeatures2TermFreq()) \
         .link(TFIDF(df_model)) \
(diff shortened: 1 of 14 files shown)


sourced/ml/transformers/indexer.py
```diff
@@ -53,21 +56,30 @@ def values(self):
         return arr
 
     def calculate_value_to_index(self, rdd: RDD):
-        column_name = self.column
-        if isinstance(column_name, str):
-            column = rdd.map(lambda x: getattr(x, column_name))
+        if self.pickle_index and os.path.exists(self.pickle_index):
+            self._log.info("Unpickling the index from %s", self.pickle_index)
+            with open(self.pickle_index, 'rb') as _in:
```
vmarkovtsev: "rb" (quotes)

sourced/ml/transformers/indexer.py
```diff
@@ -53,21 +56,30 @@ def values(self):
...
+            self._log.info("Collecting the list of distinct sorted values (%s)", column_name)
+            values = column.distinct()
+            if self.explained:
+                self._log.info("toDebugString():\n%s", values.toDebugString().decode())
+            values = values.collect()
+            values.sort()  # We do not expect an extraordinary number of distinct values
+            self._log.info("%d distinct values", len(values))
+            if len(values) == 0:
+                raise RuntimeError("Number of distinct values is zero.")
+            self._value_to_index = {d: i for i, d in enumerate(values)}
```
vmarkovtsev: The data scheme is the same as in DocumentFrequencies model which is faster and more efficient than pickling. I suggest to use it instead.

```python
DocumentFrequencies().construct(self._value_to_index).save("...")
```

sourced/ml/transformers/indexer.py
```diff
@@ -12,14 +14,15 @@ class Indexer(Transformer):
     enumerating. Use value_to_index or [] to get index value.
     """
     def __init__(self, column: Union[int, str], column2id: Union[Dict[str, int], None]=None,
-                 **kwargs):
+                 pickle_index: str=None, **kwargs):
```
vmarkovtsev: Better to name it "cached_index_path"

sourced/ml/cmd/repos2bow.py
```diff
@@ -11,21 +11,22 @@
 
 
 @pause
-def repos2bow_template(args, select=HeadFiles, cache_hook=None, save_hook=None):
+def repos2bow_template(args, select=HeadFiles, cache_hook=None, save_hook=None,
+                       full_pipeline=True):
```
vmarkovtsev: This becomes too complex. We need to split this into two functions: one can be `index_bag_features` and the second `repos2bow_template`. The second calls the first.

sourced/ml/cmd/repos2bow.py
```diff
@@ -11,21 +11,22 @@
...
     log = logging.getLogger("repos2bow")
     extractors = create_extractors_from_args(args)
     session_name = "repos2bow-%s" % uuid4()
     root, start_point = create_uast_source(args, session_name, select=select)
     uast_extractor = start_point.link(Moder(args.mode)) \
         .link(Repartitioner.maybe(args.partitions, args.shuffle)) \
         .link(Cacher.maybe(args.persist))
-    if cache_hook is not None:
+    if cache_hook is not None and full_pipeline:
```
vmarkovtsev: This will go away then: the cache hook is `None` by default in `index_bag_features`

r0mainK:
@vmarkovtsev so created a new command `repos2bow_index` and associated template for the `select`, in the end i do things differently then from `repos2bow` so i dont call one it from there. There is still a bit of duplication, but not so much. I also added a method the quantization model so we can load pre calculated levels in the extractor - i forgot about it the first time around.

sourced/ml/transformers/indexer.py
```diff
@@ -69,6 +75,11 @@ def calculate_value_to_index(self, rdd: RDD):
             raise RuntimeError("Number of distinct values is zero.")
         self._value_to_index = {d: i for i, d in enumerate(values)}
 
+    def save_index(self, path):
+        self._log.info("Saving the index to %s", self.path)
```
vmarkovtsev: `self.path` is invalid. This means that this method is not covered with tests :)

sourced/ml/models/quant.py
```diff
@@ -47,3 +47,11 @@ def dump(self):
         return """Schemes: %s""" % (
             sorted((v[0], "%d@%d" % (len(v[1]), len(next(iter(v[1].values()))) - 1))
                    for v in self.levels.items()))
+
+    def apply_quantization(self, extractors):
+        for extractor in extractors:
+            try:
+                extractor.quantize
```
vmarkovtsev: Did you forget `()`?

sourced/ml/transformers/indexer.py
```diff
@@ -12,14 +14,18 @@ class Indexer(Transformer):
...
         :param column: column index or its name in pyspark.RDD for indexing.
         :param kwargs:
         """
         super().__init__(**kwargs)
         self.column = column
         self._value_to_index = column2id
+        self.cached_index_path = cached_index_path
+        if self.cached_index_path:
+            self._log.info("Loading the index from %s", self.cached_index_path)
+            self._value_to_index = DocumentFrequencies().load(source=self.cached_index_path).df
```
vmarkovtsev: You can remove `.df` and everything is likely to continue to work. This also means that you can remove the new method in the model.

sourced/ml/models/quant.py
r0mainK: no, the method `quantize` is only defined for the `children` extractor, so it raises this error for the other extractors even if it isn't called:

```
>>> from sourced.ml.extractors import GraphletBagExtractor
>>> GraphletBagExtractor().quantize
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: 'GraphletBagExtractor' object has no attribute 'quantize'
>>> from sourced.ml.extractors import ChildrenBagExtractor
>>> ChildrenBagExtractor().quantize
<bound method ChildrenBagExtractor.quantize of <sourced.ml.extractors.children.ChildrenBagExtractor object at 0x109f8de10>>
```

sourced/ml/transformers/indexer.py
r0mainK: oh right hadn't seen the getter, on it

sourced/ml/transformers/indexer.py
r0mainK: hehe got me there, gonna add one

r0mainK:
so in the end i added a function to create or apply the quantization, since it was also used in `repos2df`, i'm also going to add a test for the function i think

sourced/ml/models/quant.py
vmarkovtsev: I don't get why this function is needed still, it seems to have no effect

r0mainK: It's pretty convoluted, but basically for all extractors with quantization (e.g. `children` only) it creates levels that are stored in `extractor.levels`. Then, when the features are extracted for all extractors you get to this in `children` (and I guess the same would go for other extractors):

```
        if not self.uast_to_bag.levels:
            # bypass NAMESPACE
            gen = self.uast_to_bag(uast).items()
        else:
            gen = super().extract(uast)
```
That is what this function does basically, just defines the level, either by computing them or loading from another model.

Issue #105, opened 2017-10-02
Check Babelfish's server version


vmarkovtsev:
Babelfish server has recently (1.0.0 release) added the new API: [Version()](https://godoc.org/github.com/bblfsh/sdk/protocol#VersionRequest). As soon as the support is added to [bblfsh/client-python](https://github.com/bblfsh/client-python), we need to add the check against the hardcoded value here.

fulaphex:
Done with #279

PR #279, opened 2018-06-21
check bblfsh server version

fulaphex:
Implementation for #105. 
I didn't find any current bblfsh server version requirements, so I used >=2.2,<3.0.
For tests I'm creating docker containers with different version, hence I needed to add `docker` to requirements.


requirements.txt
```diff
@@ -6,4 +6,5 @@ parquet>=1.2,<2.0
 numpy>=1.14
 humanize>=0.5.0,<0.6
 pygments>=2.2.0,<3.0
-keras>=2.0,<3.0
\ No newline at end of file
+keras>=2.0,<3.0
+docker>=3.4.0
```
vmarkovtsev: It should also be added to `tests_require` in `setup.py`

sourced/ml/utils/bblfsh.py
```diff
@@ -0,0 +1,18 @@
+from bblfsh.client import BblfshClient
+from distutils.version import StrictVersion
+
+BBLFSH_VERSION_LOW = "2.2"
+BBLFSH_VERSION_HIGH = "3.0"
+
+
+def check_version(host: str = "0.0.0.0", port: str = "9432") -> bool:
```
vmarkovtsev: Our style guide removes whitespaces around "=" in default arguments

sourced/ml/utils/bblfsh.py
```diff
@@ -0,0 +1,18 @@
...
+    """
+    Check if the bblfsh server version matches module requirements.
+
+    :param host: bblfsh server host
+    :param port: bblfsh server port
+    :return: True if bblfsh version specified matches requirements
+    """
+    # get version and remove leading 'v'
+    version = StrictVersion(BblfshClient(f"{host}:{port}").version().version.lstrip("v"))
```
vmarkovtsev: "f" literal will not work on Python 3.4

sourced/ml/tests/test_bblfsh_utils.py
```diff
@@ -0,0 +1,77 @@
...
+from sourced.ml.utils.bblfsh import check_version, BBLFSH_VERSION_LOW, BBLFSH_VERSION_HIGH
+
+
+def check_bblfsh_version_support(client: docker.client.DockerClient, version: str) -> bool:
+    """
+    :param client: docker client instance
+    :param version: version of bblfshd to check
+    :return: True if version is supported, False otherwise
+    """
+    for i in range(3):
```
vmarkovtsev: This seems a bit hacky. Why three times? If you need to wait until the port becomes open, use a port check with `socket.connect`.

fulaphex: I'm picking a random port, something might be running there already.

vmarkovtsev: You can check this using socket.connect

setup.py
```diff
@@ -39,6 +39,7 @@
         "tf": ["tensorflow>=1.0,<2.0"],
         "tf_gpu": ["tensorflow-gpu>=1.0,<2.0"],
     },
+    tests_require=["docker"],
```
vmarkovtsev: Docker is rapidly updated, need to specify the versions here

requirements.txt
vmarkovtsev: Put an upper constraint

sourced/ml/tests/test_bblfsh_utils.py
```diff
@@ -0,0 +1,85 @@
...
+from sourced.ml.utils.bblfsh import check_version, BBLFSH_VERSION_LOW, BBLFSH_VERSION_HIGH
+
+
+def check_bblfsh_version_support(client: docker.client.DockerClient, version: str) -> bool:
+    """
+    :param client: docker client instance
+    :param version: version of bblfshd to check
+    :return: True if version is supported, False otherwise
+    """
+    s = socket.socket()
```
vmarkovtsev: I think it is possible to write

```python
with socket.socket() as s:
    ...
```

sourced/ml/tests/test_bblfsh_utils.py
```diff
@@ -0,0 +1,85 @@
...
+    )
+
+    assert container is not None, "failed to create bblfsh container"
+
+    for i in range(3):
+        try:
+            res = check_version(port=port)
+            break
+        except Exception as e:
+            time.sleep(5)
```
vmarkovtsev: 5 seconds is too much. 100ms and 10 efforts are better

sourced/ml/tests/test_bblfsh_utils.py
```diff
@@ -0,0 +1,85 @@
...
+            pass
+
+    container.stop()
+    container.remove()
+    return res
+
+
+class BblfshUtilsTests(unittest.TestCase):
```
vmarkovtsev: Please put the tests into a separate file

fulaphex: can I just take the `check_bblfsh_version_support` and make it a method in `BblfshUtilsTests`? I don't see the reason to split it into separate files, because that method is the test itself.

vmarkovtsev: Sorry, I thought this was a normal source code file. Ignore this pls.

Issue #280, opened 2018-06-22
change pep8 -> pycodestyle in travis CI config


fulaphex:
```
pep8 has been renamed to pycodestyle (GitHub issue #466)
Use of the pep8 tool will be removed in a future release.
Please install and use `pycodestyle` instead.
```
We should change the travis config accordingly. 

@vmarkovtsev I'm happy to do it.

fulaphex:
Since I'm touching this I'd propose to use `flake8` instead of `pep8/pycodestyle`. It manages to catch a lot of unused imports and variables and in general is more extensible:
```
./sourced/ml/cmd/repos2df.py:1:1: F401 'os' imported but unused
./sourced/ml/cmd/repos2id_distance.py:5:1: F401 'sourced.ml.transformers.Cacher' imported but unused
./sourced/ml/algorithms/uast_id_distance.py:1:1: F401 'collections.namedtuple' imported but unused
./sourced/ml/algorithms/id_splitter/features.py:4:1: F401 'warnings' imported but unused
./sourced/ml/tests/test_coocc.py:1:1: F401 'os' imported but unused
./sourced/ml/tests/test_id2vec.py:3:1: F401 'os' imported but unused
./sourced/ml/tests/test_df_util.py:2:1: F401 'logging' imported but unused
./sourced/ml/tests/test_dump.py:3:1: F401 'os' imported but unused
./sourced/ml/tests/test_id_splitter_features.py:127:51: F841 local variable 'cm' is assigned to but never used
./sourced/ml/tests/models/test_tensorflow.py:9:9: F401 'tensorflow' imported but unused
./sourced/ml/models/model_converters/merge_df.py:2:1: F401 'logging' imported but unused
./sourced/ml/models/model_converters/base.py:4:1: F401 'pathlib.Path' imported but unused
./sourced/ml/transformers/uast2quant.py:4:1: F401 'pyspark.Row' imported but unused
./sourced/ml/transformers/tfidf.py:1:1: F401 'pyspark.RDD' imported but unused
./sourced/ml/extractors/id_sequence.py:1:1: F401 'typing.Tuple' imported but unused
```
@src-d/machine-learning looking for opinions on that proposal. If it flies I'll fix all those issues ^^

vmarkovtsev:
@fulaphex +1

zurk:
yeah, just fix them.

fulaphex:
Done.

PR #281, opened 2018-06-25
Change pep8 to flake8 and fix linter errors

fulaphex:
Implementation of #280 
`# flake8: noqa` at the top of the `__init__.py` files to get rid of the warnings about unused imports

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,10 @@ stages:
   - name: deploy
     # require any tag name to deploy
     if: tag =~ .*
+addons:
+  apt:
+    packages:
+      - libsnappy-dev
 _install: &_install
   - gimme 1.8
   - source ~/.gimme/envs/latest.env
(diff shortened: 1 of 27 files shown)


.travis.yml
```diff
@@ -11,6 +11,9 @@ stages:
   - name: deploy
     # require any tag name to deploy
     if: tag =~ .*
+before_install:
+  - sudo apt-get update
+  - sudo apt-get install -y libsnappy-dev
```
vmarkovtsev: Better to use apt add-ons - we could run it on macos too

vmarkovtsev:
It is a pity that we require `apt-get` though, it makes the builds longer and more unstable. Roughly every 10th build used to fail in the past. @fulaphex do we really have to install snappy?

fulaphex:
@vmarkovtsev the installation of `python-snappy` was failing on my fork when running on Travis. [travis log](https://travis-ci.org/fulaphex/ml/jobs/395473182)

It's no longer an issue ([new log](https://travis-ci.org/fulaphex/ml/jobs/399152537)), so I'll remove and PR the change.

vmarkovtsev:
@fulaphex I see, it is required by parquet. The reason it did not fail is because the cache was used. So nothing can be done, alas!

PR #284, opened 2018-06-27
Add language processing from parquet files

zurk:
For some of the commands (`repos2roleids`, `repos2id_distance`) we need to know the language for the file. And as usual, it is a good idea to start them after preprocessing step. So I add several things:

1. add `lang` field to a default set to save in preprocessing step.
2. add `LanguageSelector` step if you start from parquet files in `create_uast_source` function. 
3. Adopt `LanguageSelector` itself to work from parquet files.

### DIFF
--- a/sourced/ml/cmd/args.py
+++ b/sourced/ml/cmd/args.py
@@ -84,8 +84,13 @@ def add_repo2_args(my_parser: argparse.ArgumentParser, default_packages=None):
     # https://github.com/bblfsh/client-scala/issues/98 resolved
     languages = ["Java", "Python", "Go", "JavaScript", "TypeScript", "Ruby", "Bash", "Php"]
     my_parser.add_argument(
-        "-l", "--languages", nargs="+", choices=languages, default=languages,
+        "-l", "--languages", nargs="+", choices=languages,
+        default=None,  # Default value for --languages arg should be None.
+                       # Otherwise if you process parquet files without 'lang' column, you will
+                       # fail to process it with any --languages argument.
         help="The programming languages to analyse.")
+    my_parser.add_argument("--blacklist", action="store_true",
+                           help="Exclude languages from `--languages` list from analysis.")
     add_engine_args(my_parser, default_packages)
 
 
--- a/sourced/ml/cmd/args.py
+++ b/sourced/ml/cmd/args.py
@@ -90,7 +90,9 @@ def add_repo2_args(my_parser: argparse.ArgumentParser, default_packages=None):
                        # fail to process it with any --languages argument.
         help="The programming languages to analyse.")
     my_parser.add_argument("--blacklist", action="store_true",
-                           help="Exclude languages from `--languages` list from analysis.")
+                           help="Exclude the languages in --languages from the analysis "
+                                "instead of filtering by default.")
+    add_dzhigurda_arg(my_parser)
     add_engine_args(my_parser, default_packages)
 
 
--- a/sourced/ml/tests/test_language_selector.py
+++ b/sourced/ml/tests/test_language_selector.py
@@ -2,7 +2,8 @@ import unittest
 from pyspark.sql.functions import lit
 
 from sourced.ml.tests.models import PARQUET_DIR, SIVA_DIR
-from sourced.ml.transformers import LanguageSelector, Collector, Ignition, HeadFiles
+from sourced.ml.transformers import LanguageSelector, Collector, Ignition, HeadFiles, \
+    LanguageExtractor
 from sourced.ml.transformers.basic import create_parquet_loader, create_engine
 
 
@@ -14,6 +15,7 @@ class LoaderTest(unittest.TestCase):
         engine = create_engine("test", SIVA_DIR)
         res = Ignition(engine) \
             .link(HeadFiles()) \
+            .link(LanguageExtractor()) \
             .link(LanguageSelector(languages1)) \
             .link(Collector()) \
             .execute()
@@ -21,6 +23,7 @@ class LoaderTest(unittest.TestCase):
 
         res = Ignition(engine) \
             .link(HeadFiles()) \
+            .link(LanguageExtractor()) \
             .link(LanguageSelector(languages2)) \
             .link(Collector()) \
             .execute()
@@ -28,6 +31,7 @@ class LoaderTest(unittest.TestCase):
 
         res = Ignition(engine) \
             .link(HeadFiles()) \
+            .link(LanguageExtractor()) \
             .link(LanguageSelector(languages2, blacklist=True)) \
             .link(Collector()) \
             .execute()
@@ -35,6 +39,7 @@ class LoaderTest(unittest.TestCase):
 
         res = Ignition(engine) \
             .link(HeadFiles()) \
+            .link(LanguageExtractor()) \
             .link(LanguageSelector([])) \
             .link(Collector()) \
             .execute()
--- a/sourced/ml/tests/test_language_selector.py
+++ b/sourced/ml/tests/test_language_selector.py
@@ -1,0 +1,56 @@
+import unittest
+from pyspark.sql.functions import lit
+
+from sourced.ml.tests.models import PARQUET_DIR, SIVA_DIR
+from sourced.ml.transformers import LanguageSelector, Collector, Ignition, HeadFiles
+from sourced.ml.transformers.basic import create_parquet_loader, create_engine
+
+
+class LoaderTest(unittest.TestCase):
+    def test_parquet(self):
+        languages1 = ["Python", "Java"]
+        languages2 = ["Java"]
+
+        engine = create_engine("test", SIVA_DIR)
+        res = Ignition(engine) \
+            .link(HeadFiles()) \
+            .link(LanguageSelector(languages1)) \
+            .link(Collector()) \
+            .execute()
+        self.assertEqual(set([x.lang for x in res]), set(languages1))
+
+        res = Ignition(engine) \
+            .link(HeadFiles()) \
+            .link(LanguageSelector(languages2)) \
+            .link(Collector()) \
+            .execute()
+        self.assertEqual(set([x.lang for x in res]), set(languages2))
+
+        res = Ignition(engine) \
+            .link(HeadFiles()) \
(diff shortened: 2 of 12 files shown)


sourced/ml/transformers/basic.py
```diff
@@ -283,8 +293,13 @@ def create_uast_source(args, session_name, select=HeadFiles, language_selector=N
         parquet_loader_args = filter_kwargs(args.__dict__, create_parquet_loader)
         start_point = create_parquet_loader(session_name, **parquet_loader_args)
         root = start_point
-        if extract_uast and "uast" not in [col.name for col in start_point.execute().schema]:
+        fields = [col.name for col in start_point.execute().schema]
+        if extract_uast and "uast" not in fields:
             raise ValueError("The parquet files do not contain UASTs.")
+        if language_selector is not None:
+            start_point = start_point.link(language_selector)
```
vmarkovtsev: Why do we have to allow custom `LanguageSelector` in the first place?

The code becomes very complex and even contradictory. I would remove `language_selector` completely. Everything what it does can be repeated with `args.languages`. 

`select` arg is also breaking the abstraction because this method is command level and can be used as API level.

zurk: I agree the code is really messy right now. 
We need to rewrite it somehow. 

We allow custom `LanguageSelector` because of only one case here: https://github.com/src-d/ml/blob/master/sourced/ml/cmd/repos2ids.py#L13 

Any suggestions what we can do?

Also `LanguageSelector` itself becomes complex. may be we can break it into `LanguageExtractor` and `LanguageFilter`.

vmarkovtsev: Then better to add `--blacklist` and remove the argument and the use case.

sourced/ml/transformers/basic.py
```diff
@@ -163,24 +164,21 @@ def __call__(self, head: RDD):
...
-        return filtered_by_lang
+        if not isinstance(files, BlobsDataFrame):
+            self._log.warning("Argument type is not BlobsDataFrame. "
+                              "Language classification will be skipped.")
+        elif "lang" not in files.columns:
+            files = files \
+                .dropDuplicates(("blob_id",)) \
+                .filter("is_binary = 'false'") \
+                .classify_languages()
+        return files[files.lang.isin(self.languages) != self.blacklist]
```
vmarkovtsev: Smart!

sourced/ml/transformers/basic.py
```diff
@@ -277,22 +275,24 @@ def create_parquet_loader(session_name, repositories,
     return parquet
 
 
-def create_uast_source(args, session_name, select=HeadFiles, language_selector=None,
-                       extract_uast=True):
+def create_uast_source(args, session_name, select=HeadFiles, extract_uast=True):
```
vmarkovtsev: I find it weird that the function is named `create_uast_source` and it actually has a possibility to not extract UASTs because of `extract_uast=False`. Let's extract the common part which does not depend on `extract_uast` to new function `create_file_source` and use it from `create_uast_source`.

sourced/ml/tests/test_language_selector.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+class LoaderTest(unittest.TestCase):
+    def test_parquet(self):
+        languages1 = ["Python", "Java"]
+        languages2 = ["Java"]
+
+        engine = create_engine("test", SIVA_DIR)
+        res = Ignition(engine) \
+            .link(HeadFiles()) \
```
vmarkovtsev: Looking at these frequent `link()` calls I have got a crazy offtopic idea: how about

```python
class Transformer:
    def __rshift__(self, other):
        return self.link(other)
```

and then

```python
res = Ignition(engine) >> HeadFiles() >> LanguageSelector() >> ...
```

Thus if our train is too long to fit on a single line we can easily discard all the `\`:

```python
res = (Ignition(engine) >>
       HeadFiles() >>
       LanguageSelector() >>
        ...
```

vmarkovtsev: @r0mainK @warenlg @EgorBu what's your opinion? Or shall we use `__add__` instead?

sourced/ml/cmd/args.py
```diff
@@ -87,8 +87,13 @@ def add_repo2_args(my_parser: argparse.ArgumentParser, default_packages=None):
...
     languages = ["Java", "Python", "Go", "JavaScript", "TypeScript", "Ruby", "Bash", "Php"]
     my_parser.add_argument(
-        "-l", "--languages", nargs="+", choices=languages, default=languages,
+        "-l", "--languages", nargs="+", choices=languages,
+        default=None,  # Default value for --languages arg should be None.
+                       # Otherwise if you process parquet files without 'lang' column, you will
+                       # fail to process it with any --languages argument.
         help="The programming languages to analyse.")
+    my_parser.add_argument("--blacklist", action="store_true",
+                           help="Exclude languages from `--languages` list from analysis.")
```
vmarkovtsev: Exclude the languages in `--languages` from the analysis instead of filtering by default.

sourced/ml/tests/test_language_selector.py
r0mainK: I like the look of it personally, however don't really like:
```
res = (Ignition(engine) >>
       HeadFiles() >>
       LanguageSelector() \
       .execute()
```

warenlg: it looks nice to me also +1

vmarkovtsev: @r0mainK We can go wild and
```python
class Transformer:
    def __inv__(self):
        return self.execute()
```

```python
res = ~(Ignition(engine) >>
       HeadFiles() >>
       LanguageSelector())
```

Now it starts to look like real cryptic C++!

zurk: Nice!
Actually, we can have a 
```python
class Execute(Transformer):
    def __call__(self, head):
        return head.execute()
```

So 
```python
res = (Ignition(engine) >>
        HeadFiles() >>
        LanguageSelector() >>
        Execute())
```

I just do not like to use inverse symbol `~` for it.

vmarkovtsev: Good idea

vmarkovtsev: Who wants to implement?

zurk: I will in another PR

zurk:
@vmarkovtsev I really like your suggestion about `create_uast_source` splitting, so I dig a little further and remove `select` argument, because it is always `HeadFiles` or `DzhigurdaFiles` and `HeadFiles()` equal to `DzhigurdaFiles(dzhigurda=0)`.

and one more improvement with `UastExtractor` to avoid conditioning.

zurk:
sourced-ml becomes nicer and nicer.

vmarkovtsev:
@zurk ping here

sourced/ml/transformers/basic.py
```diff
@@ -163,33 +166,45 @@ def __call__(self, head: RDD):
...
 
 
 class UastExtractor(Transformer):
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
 
     def __call__(self, files: DataFrame) -> DataFrame:
-        # if UAST is not extracted, returns an empty list that we filter out here
-        return files.extract_uasts().where(functions.size(functions.col("uast")) > 0)
+        if isinstance(files, (BlobsDataFrame, BlobsWithLanguageDataFrame)):
```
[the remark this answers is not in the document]
zurk: not really, the problem here is in parquet files. If you start with them, UAST can already be extracted. Of course, we can check it in the middle of the pipeline, but I prefer to move this check to this place because of our functions like `create_uast_source` were extremely complex, branched and confusing. Now it is only 3 lines.

sourced/ml/cmd/args.py
```diff
@@ -84,8 +84,15 @@ def add_repo2_args(my_parser: argparse.ArgumentParser, default_packages=None):
     # https://github.com/bblfsh/client-scala/issues/98 resolved
     languages = ["Java", "Python", "Go", "JavaScript", "TypeScript", "Ruby", "Bash", "Php"]
     my_parser.add_argument(
-        "-l", "--languages", nargs="+", choices=languages, default=languages,
```
vmarkovtsev: We don't need `languages` then

sourced/ml/transformers/basic.py
```diff
@@ -206,33 +209,45 @@ def __call__(self, head: RDD):
...
-            for lang in self.languages[1:]:
-                lang_filter |= classified.lang == lang
-        else:
-            lang_filter = classified.lang != self.languages[0]
-            for lang in self.languages[1:]:
-                lang_filter &= classified.lang != lang
-        filtered_by_lang = classified.filter(lang_filter)
-        return filtered_by_lang
+        if not isinstance(files, BlobsDataFrame):
+            self._log.warning("Argument type is not BlobsDataFrame. "
```
vmarkovtsev: We should print the warning only if `languages` is not None

sourced/ml/transformers/basic.py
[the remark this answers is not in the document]
vmarkovtsev: Refactor, nuke, rewrite but I need UastExtractor to extract uasts and accept the single data type RDD. The rest goes to the upper scope.

sourced/ml/transformers/basic.py
zurk: it is an error if `languages is None` here. because we do 
`return files[files.lang.isin(self.languages) != self.blacklist]`
hard to tell what we expect to happen if `self.languages is None`.
So It can not be None. We use `LanguageSelector` only in one place: https://github.com/src-d/ml/pull/284/files#diff-370750368c8314a487dfbae4ca854901R347 and with `.maybe()`. it checks` if languages is None`

vmarkovtsev: Then you must type assert for `BlobsDataFrame` (or `BlobsWithLanguageDataFrame`?) instead of the warning. The input type must be exactly one possible class and it is the responsibility of the caller to ensure that.

Issue #270, opened 2018-06-14
Fix assemble_spark_config that alter default values now


zurk:
I found out in tests that `SparkDefault.PACKAGES` and `SparkDefault.CONFIG` were changed. 
To fix it I need to change a behavior of `assemble_spark_config` function.

zurk:
@vmarkovtsev please assign me.

r0mainK:
@zurk  could we actually change `SparkDefault.PACKAGES` to `SparkDefault.JARS` ? because packages could be used to manage python package dependencies that are not on the workers, like we do for `sourced-ml` and `sourced-engine`.

zurk:
I am not sure it is a good idea. It is the same as name variable for python packages as `PY`. And I do not see any conflict in naming here.

If you really what we can choose something like `SparkDefault.JAR_PACKAGES`. It corresponds to `spark.jars.packages` config variable in spark.

r0mainK:
True, there is no naming conflicts (yet), but I think it would be more verbose then simply `package`. Anyway Im all for `SparkDefault.JAR_PACKAGES` we'll just have to port it apollo afterwards

zurk:
Done

PR #287, opened 2018-06-29
Fix assemble_spark_config default values altering

zurk:
this PR closes issue https://github.com/src-d/ml/issues/270.
I change the behavior of functions that assembles configs for spark and engine. It was found out that they change default value variables. Now, default values are tuples and can not be changed.

Also, I add default values for missing cases and polish code.


sourced/ml/utils/engine.py
```diff
@@ -23,56 +43,41 @@ class Columns:
...
-        engine = get_distribution("sourced-engine").version
-    except DistributionNotFound:
-        log = logging.getLogger("engine_version")
-        engine = requests.get("https://api.github.com/repos/src-d/engine/releases/latest") \
-            .json()["tag_name"].replace("v", "")
-        log.warning("Engine not found, queried GitHub to get the latest release tag (%s)",
-                    engine)
-    return engine
-
+def create_engine(session_name, repositories,
```
vmarkovtsev: Would be nice to have a docstring which explains these arguments

PR #288, opened 2018-07-04
Update modelforge requirements

r0mainK:
all in title


r0mainK:
@vmarkovtsev so tests are failing bc of changes in modelforge, I would need to modify modelforge a bit (basically move arg parsers out of the main ) to icorrect errors cleanly.
However, why do we have this dump in commands in the first place ? It does exactly the same as modelforge with less features, and modelforge is a requirement of ml. IMO we should take it out.

vmarkovtsev:
Looks fair to remove.

r0mainK:
@vmarkovtsev dump tests seem to be failing due to fkin dulwich: basically `git clone` fails. I added skip for the DumpTest with a TODO. Alternatively, since `dump` does not need the index if there is no backend, I could simply modify modelforge, so it's as you want.

r0mainK:
@vmarkovtsev to be merged when you can

vmarkovtsev:
@r0mainK Expecting a better `git clone` workaround/fix than `unittest.skip` :wink:

PR #291, opened 2018-07-09
Dulwich is fixed !

r0mainK:
All in title, took out the skip unittest and removed unnecessary parts

### DIFF
--- a/sourced/ml/tests/test_dump.py
+++ b/sourced/ml/tests/test_dump.py
@@ -26,7 +26,6 @@ def captured_output():
         logging.getLogger().removeHandler(log_handler)
 
 
-@unittest.skip  # TODO: when dulwich is fixed, update
 class DumpTests(unittest.TestCase):
     ID2VEC_DUMP = """{'created_at': datetime.datetime(2017, 6, 18, 17, 37, 6, 255615),
  'dependencies': [],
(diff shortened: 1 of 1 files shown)


r0mainK:
@vmarkovtsev can be merged

sourced/ml/tests/test_dump.py
```diff
@@ -77,38 +76,26 @@ def tearDown(self):
...
             shutil.rmtree("/tmp/ml-test-dump")
 
-    def setUp(self):
-        self.maxDiff = None
-
     def test_id2vec(self):
-        with captured_output() as (out, err, log):
+        with captured_output() as (out, _, _):
             dump_model(self._get_args(input=paths.ID2VEC))
-        self.assertEqual(err.getvalue(), "")
```
zurk: No reason to remove it. It checks that `stderr` is empty and it should be.
Same for `log` Here and everywhere.

r0mainK: (Discussed this in real time, backloging):
So I added this for debugging purpose in previous PR for modelforge requirements. Removed because: 
- tests is meant to check ours model's `dump` works, that's it
- dulwich still gives off random (but apparently non breaking) errors in `stderr` which makes build fail

PR #293, opened 2018-07-10
Add __rshift__ __lshift__ shortcuts for link and unlink transformers

zurk:
As was discussed here: https://github.com/src-d/ml/pull/284#discussion_r200116805 we decided to add code that will allow us to write  pipelines like:
```python
Ignition(engine) >> HeadFiles() >> LanguageSelector() >> Execute()
```

This PR contains main changes to allow this. Note, that I also add `<<` for `unlink()` method.

Also, I modify `link()` function to allow multiple likage. Example:
```
t1 = DumpTransformer(1)
t2 = DumpTransformer(2)
t31 = DumpTransformer(31)
t32 = DumpTransformer(32)

t1 >> t2 >> (t31, t32)
t1 >> Explode()
```
It creates and executes a tree:
```
t1 - t2 - t31
      \
       t32
```

To be able to use `>> Execute()` and `>> Explode()` I create special `LeafTransformer`s that cannot be connected to anything else and marks that it is time to execute pipeline. I think it is a most simple way to implement it.

As soon as this PR will be merged, I plan to rewrite our pipelines via `>>`.

### DIFF
--- a/sourced/ml/transformers/transformer.py
+++ b/sourced/ml/transformers/transformer.py
@@ -1,5 +1,5 @@
 from io import StringIO
-from typing import Iterable, Union, List
+from typing import Union, Tuple
 
 from sourced.ml.utils import PickleableLogger  # nopep8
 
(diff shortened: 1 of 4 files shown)


sourced/ml/transformers/transformer.py
```diff
@@ -128,3 +143,29 @@ def _format_pipeline(pipeline):
...
+        raise Exception("It is not possible to link anything to LeafTransformer.")
+
+
+class Execute(LeafTransformer):
+    """
+    Special transformer to execute all the pipeline.
+    As soon as one links anything to this Transformer it call execute() for the pipeline.
+    It is not possible to link anything to this transformer.
+    """
+
```
r0mainK: think you forgot a pass here ?

zurk: yes

sourced/ml/transformers/transformer.py
```diff
@@ -41,7 +42,13 @@ def path(self):
         path.reverse()
         return path
 
-    def link(self, transformer):
+    def link(self, transformer: Union[Iterable, "Transformer"]) -> Union[List, "Transformer"]:
```
vmarkovtsev: Why not `def link(self, *transformer: "Transformer")`? Also better

```python
-> Union[Tuple["Transformer"], "Transformer"]
```

sourced/ml/transformers/transformer.py
```diff
@@ -41,7 +42,13 @@ def path(self):
...
+        if isinstance(transformer, Execute):
+            return self.execute()
+        if isinstance(transformer, Explode):
+            return self.explode()
```
vmarkovtsev: I am sure we can do the same without these super hacky conditions.

PR #295, opened 2018-07-11
Add merge bow function

r0mainK:
We have command for merging `docfreq` models, this implements the same thing for `bow` models. I think it is useful because on large inputs of data we can't collect all of them at once and create small batches of bags due to Spark limitations, however it is cumbersome to move around dozen or hundreds of those, and not compact. I also implemented a function to reduce the output bag by keeping only user selected features (default is all). I did that because it ill probably be useful for users, especially if they are downloading the model but only want specific(s) features, however not sure whether it shouldn't also be a method of the BOW model itself, if yes tell me and I'll change structure. I also removed some unused stuff that we forgot to take out.

### DIFF
--- a/sourced/ml/models/model_converters/merge_bow.py
+++ b/sourced/ml/models/model_converters/merge_bow.py
@@ -1,0 +1,68 @@
+import os
+from scipy.sparse import vstack
+
+from sourced.ml import extractors
+from sourced.ml.models.bow import BOW
+from sourced.ml.models.model_converters.base import Model2Base
+
+
+class MergeBOW(Model2Base):
+    """
+    Merges several :class:`BOW` models together.
+    """
+    MODEL_FROM_CLASS = BOW
+    MODEL_TO_CLASS = BOW
+
+    def __init__(self, features=None, *args, **kwargs):
+        super().__init__(num_processes=1, *args, **kwargs)
+        self.documents = None
+        self.tokens = None
+        self.matrix = None
+        self.deps = None
+        self.features_namespaces = None
+        if features:
+            self.features_namespaces = [ex.NAMESPACE for ex in extractors.__extractors__.values()
+                                        if ex.NAME in features]
+
+    def convert_model(self, model: BOW) -> None:
+        if self.tokens is None:
+            self.tokens = model.tokens
+            self.documents = model.documents
+            self.matrix = [model.matrix.tocsr()]
+            self.deps = model._meta["dependencies"]
+        elif set(self.tokens) != set(model.tokens):
+            raise ValueError("Models don't share the same set of tokens !")
+        else:
+            self.documents += model.documents
+            self.matrix.append(model.matrix.tocsr())
+
+    def finalize(self, index: int, destdir: str):
+        self._log.info("Stacking matrices ...")
+        matrix = self.matrix.pop(0)
+        while self.matrix:
+            matrix = vstack([matrix, self.matrix.pop(0)])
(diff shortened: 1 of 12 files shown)


sourced/ml/models/model_converters/merge_bow.py
```diff
@@ -0,0 +1,69 @@
...
+from scipy.sparse import hstack, vstack
+
+from sourced.ml import extractors
+from sourced.ml.models.bow import BOW
+from sourced.ml.models.model_converters.base import Model2Base
+
+
+class MergeBOW(Model2Base):
+    """
+    Merges several :class:`DocumentFrequencies` models together.
```
zurk: Merges several :class:~`DocumentFrequencies`~ BOW  models together.

sourced/ml/models/model_converters/merge_bow.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def __init__(self, features=None, *args, **kwargs):
+        super().__init__(num_processes=1, *args, **kwargs)
+        self.documents = None
+        self.tokens = None
+        self.matrix = None
+        self.deps = None
+        self.features = None
+        if features:
+            self.features = [ex.NAMESPACE for ex in extractors.__extractors__.values()
```
zurk: better to rename to `features_namespaces`
otherwise, it confuses names and namespaces

sourced/ml/models/model_converters/merge_bow.py
```diff
@@ -0,0 +1,69 @@
...
+        elif set(self.tokens) != set(model.tokens):
+            raise ValueError("Models don't share the same set of tokens !")
+        else:
+            self.documents += model.documents
+            self.matrix.append(model.matrix.tocsr())
+
+    def finalize(self, index: int, destdir: str):
+        self._log.info("Stacking matrices ...")
+        matrix = self.matrix[0]
+        self.matrix.pop(0)
```
zurk: `matrix = self.matrix.pop(0)`

sourced/ml/models/model_converters/merge_bow.py
```diff
@@ -0,0 +1,69 @@
...
+        while self.matrix:
+            matrix = vstack([matrix, self.matrix[0]])
+            del self.matrix[0]
```
zurk: `matrix = vstack([matrix, self.matrix.pop(0)]])`

sourced/ml/models/model_converters/merge_bow.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def _reduce_matrix(self):
+        reduced_tokens = []
+        columns = []
+        for i, token in enumerate(self.tokens):
+            if token.split(".")[0] in self.features:
+                reduced_tokens.append(token)
+                columns.append(self.matrix.getcol(i).tocsc())
+        self.tokens = reduced_tokens
+        self.matrix = hstack(columns)
```
zurk: I think It is better to convert `self.matrix` to csc first 
(`matrix = self.matrix.tocsc()`), 
than in loop  ~`columns.append(self.matrix.getcol(i).tocsc())`~ `columns.append(i)` 
and here ~`self.matrix = hstack(columns)`~ `self.matrix = matrix[columns]`

sourced/ml/models/model_converters/merge_bow.py
r0mainK: agreed

PR #296, opened 2018-07-11
Split BagsExtractor into Extractor and BagsExtractor itself

zurk:
There are already several cases and definitely will be more when we need to process UAST and get some dataset, but as output do not have a BOW model. But we have this fancy pipeline like:
```python
start_point \
        .link(UastRow2Document()) \
        .link(UastDeserializer()) \
        .link(Uast2BagFeatures([extractor]))
```
And it is reasonable to have something similar in case of not-bag data. Before this PR you can actually use the `Uast2BagFeatures` transformer for it. However, some ad-hoc code should be applied to make your `BagsExtracor` work in this case. For example, ~Bag~Extractor to collect Role-Identifiers dataset. 

So I decided to split `BagsExtracor` into `Extractor` and `BagsExtracor` itself.
 
## Changes list
1. `BagsExtractor` splits into `Extractor` and `BagsExtracor`. `BagsExtracor` are mostly the same. `Extractor` is a general wrapper for any UAST algorithm we want to use. As an example see `3.`
2. `Uast2BagFeatures` splited to `Uast2Features` and `Uast2BagFeatures` itself. The first one because now it works with `BagsExtracor`. `Uast2BagFeatures` now it is super simple but produces the same result as before. 
3. `RolesAndIdsExtractor` was completely removed because it was ad-hoc code. In the next PRs, I plan to rewrite other ad-hoc `Extractors`.

This is the main changes to our codebase. As soon as this PR merged, I refactor rest code to use new classes where it is necessary. Also, `BagsExtractor` will use `algorithm` field. now it is omitted on purpose


sourced/ml/extractors/bags_extractor.py
```diff
@@ -18,7 +51,7 @@ def __init__(self, docfreq_threshold=None, weight=None, **kwargs):
         :param weight: TF-IDF will be multiplied by this weight to change importance of specific \
                       bag extractor
         """
-        super().__init__(**kwargs)
+        super().__init__(algorithm=None, **kwargs)
```
vmarkovtsev: thus this should go away

sourced/ml/extractors/bags_extractor.py
```diff
@@ -1,14 +1,47 @@
...
+    Converts a single UAST via `algorithm` to anything you need.
+    It is a wrapper to use in `Uast2Features` Transformer in a pipeline.
+    """
+    NAME = None  # feature scheme name, should be overridden in the derived class.
+    OPTS = dict()  # cmdline args which are passed into __init__()
+
+    def __init__(self, algorithm: Callable, name: Optional[str]=None, **kwargs):
+        super().__init__(**kwargs)
+        if name is not None:
+            self.NAME = name
```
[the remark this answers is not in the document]
vmarkovtsev: You must rely on the abstract `extract()` and remove the constructor.

sourced/ml/transformers/uast2bag_features.py
```diff
@@ -28,7 +28,27 @@ def documentize(self, r: Row) -> Row:
         return Row(**{bfc.document: doc, ec.Uast: r[ec.Uast]})
 
 
-class Uast2BagFeatures(Transformer):
+class Uast2Features(Transformer):
+    def __init__(self, extractors: Union[Extractor, Iterable[Extractor]], **kwargs):
```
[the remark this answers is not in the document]
zurk: good idea!

sourced/ml/extractors/bags_extractor.py
```diff
@@ -1,14 +1,44 @@
...
+class Extractor(PickleableLogger):
+    """
+    Converts a single UAST via `algorithm` to anything you need.
+    It is a wrapper to use in `Uast2Features` Transformer in a pipeline.
+    """
+    NAME = None  # feature scheme name, should be overridden in the derived class.
+    ALGORITHM = None  # algorithm class to extract from UAST
+    OPTS = dict()  # cmdline args which are passed into __init__()
+
+    def __init__(self, **kwargs):
```
vmarkovtsev: You can remove this empty __init__

sourced/ml/extractors/bags_extractor.py
```diff
@@ -1,14 +1,44 @@
...
+    def get_kwargs_fromcmdline(cls, args):
+        prefix = cls.NAME + "_"
+        result = {}
+        for k, v in args.__dict__.items():
+            if k.startswith(prefix):
+                result[k[len(prefix):]] = v
+        return result
+
+    def extract(self, uast: bblfsh.Node):
+        yield from self.ALGORITHM(uast)
```
vmarkovtsev: Good :+1:

sourced/ml/extractors/bags_extractor.py
vmarkovtsev: All right, this must be updated for `ALGORITHM`

sourced/ml/transformers/uast2bag_features.py
```diff
@@ -28,26 +28,34 @@ def documentize(self, r: Row) -> Row:
...
-    def __init__(self, extractors: Iterable[BagsExtractor], **kwargs):
+class Uast2Features(Transformer):
+    def __init__(self, *extractors: Extractor, **kwargs):
         super().__init__(**kwargs)
         self.extractors = extractors
 
     def __call__(self, rows: RDD):
         return rows.flatMap(self.process_row)
 
+    def process_feature(self, row: Row, name, feature):
```
vmarkovtsev: This must be implemented in the derived class.

UastMiner -> Uast2FeaturesDict, Uast2BagOfFeatures

(I am open to better names)

zurk: UastMiner, Uast2Features, Uast2BagFeatures?

Issue #298, opened 2018-07-16
Add "single-shot" mode to the token parser


vmarkovtsev:
Our heuristic token parser generates several variants of the split sometimes. E.g.
`wdSize` becomes `["wd", "size", "wdsize"]`.

We need to add an "aggressive" mode which does not generate repetitions and use it to train our neural identifier splitter. Besides, I want the following in this mode:

```
loooong_sh_loooong_sh -> looong, sh, loooong, sh
SmallIdFooo -> small, id, foo
```

zurk:
Yes, because of https://github.com/src-d/ml/pull/204
On my way

zurk:
Addressed. 

Going to fix the dataset.

zurk:
Actually, we can use `split` instead of `process_token` as @warenlg did here: https://github.com/src-d/ml/blob/master/sourced/ml/transformers/content2ids.py#L67 it gives us exactly the same result you want. So, I am not sure this PR is helpful. It just makes things more complicated.

vmarkovtsev:
But wait I see that you've changed the code to not return duplicate tokens. Waren's code suffers from the same problem.

zurk:
Ok. It was a too fast conclusion. But in case you set `min_split_length=1` you will have only independent part of an identifier. Because only small pieces sticks.

zurk:
PR is merged. 
@vmarkovtsev do not forget to close :)

PR #299, opened 2018-07-16
Feature/one shot

zurk:
Related issue: https://github.com/src-d/ml/issues/298


sourced/ml/algorithms/token_parser.py
```diff
@@ -11,14 +11,19 @@ class TokenParser:
     STEM_THRESHOLD = 6  #: We do not stem splitted parts shorter than or equal to this size.
     MAX_TOKEN_LENGTH = 256  #: We cut identifiers longer than this value.
     MIN_SPLIT_LENGTH = 3  #: We do not split source code identifiers shorter than this value.
+    DEFAULT_SINGLE_SHOT = True  #: True if we do not want to join small identifiers to next one.
```
vmarkovtsev: Should be False by default to preserve the old behavior

zurk:
Just note, that I include your commit also (PR https://github.com/src-d/ml/pull/297)

PR #301, opened 2018-07-25
Improve coverage and make tests UAST v2 compatible

warenlg:


sourced/ml/transformers/basic.py
```diff
@@ -289,17 +289,20 @@ def __call__(self, rows: RDD) -> RDD:
...
+            row_dict[EngineConstants.Columns.Uast] = []
+            for i, uast in enumerate(row[EngineConstants.Columns.Uast]):
+                try:
+                    row_dict[EngineConstants.Columns.Uast].append(self.parse_uast(uast))
+                except:  # noqa
+                    self._log.error("\nBabelfish Error: Failed to parse uast #%s for repository "
+                                    "%s" % (i, row[Uast2BagFeatures.Columns.repository_id]))
+            yield Row(**row_dict)
+        except ValueError:
+            self._log.error("The RDD provided does not include uasts.")
```
vmarkovtsev: `if not row[EngineConstants.Columns.Uast]:` must check exactly this. If it is not enough, change that lane please

vmarkovtsev:
@warenlg Please resolve the conflicts

vmarkovtsev:
@zurk Wlll bump the version after your review and merge here.

sourced/ml/tests/test_basic_transformers.py
```diff
@@ -51,14 +90,88 @@ def test_identity(self):
...
+            for f in files:
+                filename = os.path.join(root, f)
+                if filename.endswith(".csv"):
+                    with open(filename) as f:
+                        reader = csv.reader(f)
+                        next(reader)
+                        data = [r for r in reader]
+
+        self.assertEqual(len(data), 1)
+        self.assertEqual(data[0][0], rows[0][0])
```
zurk: and
```python
self.assertEqual(data[0][1], rows[0][1])
```

sourced/ml/transformers/basic.py
```diff
@@ -30,12 +30,12 @@ def __call__(self, head: RDD):
         if self.keymap is None:
             return head.coalesce(self.partitions, self.shuffle)
         # partitionBy the key extracted using self.keymap
-        if self.keymap is not False:
+        else:
```
zurk: is it a bug fix?

sourced/ml/transformers/basic.py
```diff
@@ -300,16 +300,16 @@ def __call__(self, rows: RDD) -> RDD:
         return rows.flatMap(self.deserialize_uast)
 
     def deserialize_uast(self, row: Row):
-        if not row[EngineConstants.Columns.Uast]:
+        if EngineConstants.Columns.Uast not in row:
             return
```
zurk: I think we should keep both conditions because `row[EngineConstants.Columns.Uast]` can be an empty list and we do not want an empty list further:
```python
if EngineConstants.Columns.Uast not in row or not row[EngineConstants.Columns.Uast]:
    return
```

sourced/ml/transformers/basic.py
```diff
@@ -300,16 +300,16 @@ def __call__(self, rows: RDD) -> RDD:
...
         row_dict = row.asDict()
         row_dict[EngineConstants.Columns.Uast] = []
         for i, uast in enumerate(row[EngineConstants.Columns.Uast]):
             try:
                 row_dict[EngineConstants.Columns.Uast].append(self.parse_uast(uast))
             except:  # noqa
-                self._log.error("\nBabelfish Error: Failed to parse uast for document %s for uast "
-                                "#%s" % (row[Uast2BagFeatures.Columns.document], i))
+                self._log.error("\nBabelfish Error: Failed to parse uast #%s for repository "
+                                "%s" % (i, row[Uast2BagFeatures.Columns.repository_id]))
```
zurk: it is always better to have `document` name if it is available. Easier to report a bug. If it is not available we can use `repository_id ` of course. 

Usually, we have  `document` name because we use `UastRow2Document` transformer.

sourced/ml/transformers/basic.py
warenlg: I did this because the code returned a ValueError when the dataframe did not contain the "uast" column https://github.com/src-d/ml/pull/301#discussion_r212960068
Do you want this : `if EngineConstants.Columns.Uast not in row and not row[EngineConstants.Columns.Uast]` ?

sourced/ml/transformers/basic.py
warenlg: No, I just thought it was more readable like this

sourced/ml/transformers/basic.py
zurk: ok, let's put both conditions, I am ok with that.

warenlg:
@vmarkovtsev @zurk good for you ?

sourced/ml/transformers/basic.py
```diff
@@ -32,10 +32,10 @@ def __call__(self, head: RDD):
...
         if self.keymap is not False:
             # user knows what they are doing
-            head = head.map(lambda x: (self.keymap(x), x))
-        return head \
-            .partitionBy(self.partitions) \
-            .map(lambda x: x[1])
+            return head \
+                .map(lambda x: (self.keymap(x), x)) \
+                .partitionBy(self.partitions) \
+                .map(lambda x: x[1])
```
vmarkovtsev: So if `self.keymap is False` we return nothing.
None is not the same as False. But damn the whole logic here is twisted. Can ask the code author what we should do here and also add the comment which explains that `keymap` can be False, None or real callable.
Personally, I think that having a difference between False and None is very confusing.

zurk: Yeah, I agree it is very confusing. But there is a comment in class __docstring__:
```python
"""
     Repartitioner uses one of the three ways to split an RDD into partitions:
     1. coalesce() if shuffle=False, keymap=None
     2. repartition() if shuffle=True, keymap=None
     3. partitionBy() if keymap is not None
"""
```

I think we can separate 3rd case into separate Transformer.

vmarkovtsev: No need to separate and the comment is good.

The confusion comes from different actions taken when keymap is None and keymap is False.

zurk: ok, now I get it. @warenlg now I noticed that you changed the logic of this code. 
Now it is
```python
    def __call__(self, head: RDD):
        if self.keymap is None:
            return head.coalesce(self.partitions, self.shuffle)
        # partitionBy the key extracted using self.keymap
        if self.keymap is not False:
            # user knows what they are doing
            return head \
                .map(lambda x: (self.keymap(x), x)) \
                .partitionBy(self.partitions) \
                .map(lambda x: x[1])
```
and it was
```python
    def __call__(self, head: RDD):
        if self.keymap is None:
            return head.coalesce(self.partitions, self.shuffle)
        # partitionBy the key extracted using self.keymap
        if self.keymap is not False:
            # user knows what they are doing
            head = head.map(lambda x: (self.keymap(x), x))
        return head \
            .partitionBy(self.partitions) \
            .map(lambda x: x[1])
```

So in case `self.keymap is False` you return nothing and it is an error. if `self.keymap is False` transformer should just skip mapping step.
I think it is good idea to cover this option with additional test.

vmarkovtsev:
I resolved myself, waiting for CI results.

vmarkovtsev:
CI passed

PR #309, opened 2018-08-28
Polish id2role_eval code, remove the trash

zurk:
Following after https://github.com/src-d/ml/pull/308

Changes:
1. Delete random comments
1. Evaluate on test dataset for better generalization ability estimation. 🤦‍♂️

PR #310, opened 2018-08-28
Bump version

zurk:
It is time to do it. I want to update Apollo requirements.
I prefer to have https://github.com/src-d/ml/pull/309 merged before.
@vmarkovtsev We also should add a tag 0.6.0


zurk:
ok, now we can deal with this one: https://github.com/src-d/apollo/pull/66

PR #324, opened 2018-10-06
setup.py: Add long_description and python_requires

hugovk:
The description is missing at https://pypi.org/project/sourced-ml/:

image

This PR adds it, using README.md. See https://dustingram.com/articles/2018/03/16/markdown-descriptions-on-pypi for instructions, the main thing is to make sure to use recent twine and setuptools for upload: `pip install -U twine setuptools`.

---

Also, add `python_requires` to help pip install the right version for the user's Python version.
 
More info on how this works:
* https://packaging.python.org/guides/dropping-older-python-versions/
* https://hackernoon.com/phasing-out-python-runtimes-gracefully-956f112f33c4

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -47,6 +47,7 @@ setup(
     tests_require=["docker>=3.4.0,<4.0"],
     package_data={"": ["LICENSE.md", "README.md"],
                   "sourced": ["ml/transformers/languages.yml"], },
+    python_requires=">=3.4",
     classifiers=[
         "Development Status :: 3 - Alpha",
         "Environment :: Console",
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -14,6 +17,8 @@
     description="Framework for machine learning on source code. "
                 "Provides API and tools to train and use models based "
                 "on source code features extracted from Babelfish's UASTs.",
+    long_description=long_description,
```
vmarkovtsev: @zurk Can you please add this nice feature to all our Python projects.

zurk: sure

Issue #271, opened 2018-06-14
Create terms glossary for sourced.ml


zurk:
We constantly confuse terms, so what to say about other developers.
I do not want to make it full, but to have a start. 

Here is terms list to explain on the first iteration:
1. identifier
2. token
3. Bag-of-words
4. Weighted bag-of-words
5. Model 
6. Algorithm
7. Transformer

@src-d/machine-learning please take a look and add any confusing terms you remember.

r0mainK:
If we're gonna define identifiers and token, might as well also add literals, graphlets and also ~quantification~ quantization . I think we could divide the glossary into:
-  terms that mean something more specific then would be usually the case or are vague to start with e.g. model meaning a modelforge model, words in BOW being any feature extracted from a document, document that means a repo/file or function, etc.
-  terms that we use in the same ways it is intended but not be well known. Now of course they have Google, but we might as well drop a couple lines to explain the concept. E.g. COOC, quantization, topics, TFIDDF

vmarkovtsev:
Linking to https://github.com/src-d/apollo/blob/master/doc/GLOSSARY.md

zurk:
Thanks, @r0mainK I update the description.

PR #329, opened 2018-10-10
Add a glossary of terms to sourced.ml

eindiran:
An attempt to flesh out the glossary requested in issue #271.
Not all terms are completely filled in yet, so this PR is more or less a work-in-progress.

If I've gotten anything wrong or some description is ambiguous, if you let me know I can fix it.

Signed-off-by: eindiran <eindiran@uchicago.edu>


zurk:
Indeed! Thank you!

Issue #332, opened 2018-10-17
Avoid using pandas


warenlg:
pandas slows down a lot the installation and is used only once in the repo https://github.com/src-d/ml/blob/master/sourced/ml/cmd/id2role_eval.py#L15 I believe we could avoid using it, and use the `CSVSaver` method instead https://github.com/src-d/ml/blob/master/sourced/ml/transformers/basic.py#L72.

zurk:
yeah, I remember this one. Unfortunately, I use `pandas` more than just CSV loader here. `isin`, `groupby` is not so easy to exclude. But what is definitely can be done is to move `pandas` to `extras_require` as we do with tensorflow.

I can PR it if you want.

zurk:
Is it fixed?

PR #335, opened 2018-10-22
Move pandas from requirements to extras

warenlg:

This is only a temporary fix for https://github.com/src-d/ml/issues/332 as I'd like to remove pandas df to use python RDD in `id2role-eval`. However I'd need to write the tests for this command, so it's not a quick fix


vmarkovtsev:
Result: it imports.

PR #344, opened 2018-10-31
Delete hiding sourced-engine usage from algorithms

zurk:
We should make a new release after the merge.

### DIFF
--- a/sourced/ml/__init__.py
+++ b/sourced/ml/__init__.py
@@ -5,4 +5,4 @@ try:
 except ImportError:
     pass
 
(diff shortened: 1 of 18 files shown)


zurk:
@vmarkovtsev ready. A lot of imports changed. 
But the main idea is  
1. Import `from sourced.ml.utils import bblfsh_roles ` if you need roles, do not import it separately
2. exclude `sourced.ml.utils.engine` and `sourced.ml.utils.spark` from `sourced/ml/utils/__init__.py`. And if they are required, import them directly.

PR #353, opened 2018-11-19
Add style check stage

zurk:
Add style check stage in Travis.
Also, harmonize style check with https://github.com/src-d/style-analyzer.

Better to review by commits.


zurk:
wow, GitHub marks force-pushes now.

vmarkovtsev:
Yep, did not see them last week...

@zurk Can you please make the single `.flake8` config the same way it works in the current `lookout-sdk-ml` or `style-analyzer`. BTW: damn, you've done much work!!

zurk:
done.

zurk:
rebased and fix `requirements.txt` conflict

zurk:
Something bad happened in dependencies for Python 3.4. Maybe it is a good time to drop it, @vmarkovtsev?

vmarkovtsev:
@zurk Let's return to this problem when we migrate to UASTv2 - the new client does not support 3.4.

This particular problem is because pip updated to version 19 and broke quite many things. We need to pin its version to 18.1 (or even older if the problem persisits). Here is how: https://github.com/src-d/hercules/blob/master/.travis.yml#L51

PR #399, opened 2019-03-29
Fix a missing argument (series) to the call of the save function

Jan21:
Signed-off-by: Jan Hula <jan.hula21@gmail.com>


sourced/ml/transformers/bow_writer.py
```diff
@@ -79,7 +79,7 @@ def __call__(self, head: RDD):
             filename = self.get_bow_file_name(self.filename, i)
             BOW() \
                 .construct(docs, tokens, matrix) \
-                .save(filename, deps=(self.df,))
+                .save(filename, series="Bow", deps=(self.df,))
```
vmarkovtsev: You should require "series" argument in `__init__`, because BowWriter is an implementation of writing the model and it cannot know the specific from the inside.

PR #405, opened 2019-04-16
[WIP] NN tokenizer

glimow:
@warenlg's neural network tokenizer integration into `modelforge` and src-d's `TokenParser`.
- The model interface with `modelforge` is up for review ! feel free to make some because I am not familiar with src-d's conventions yet
- The tests take still ~1min to run because of the model's size (8mb), which make it's loading from the asdf file very slow.
- Adding the model to `TokenParser` is the WIP part.


zurk:
@glimow thanks!
several things first.
1. Please use rebase inside your PR instead of merges to make everything easier to review. 
2. You should make DCO bot happy: add a signature to a commit message with `git commit -s`
3. You should make Travis happy. I see you use `/tmp/` in the tests. do not do that, we have a check :)  Use whatever you need from `import tempfile`.

to run the same style check we use in Travis run `make check`

glimow:
1- OK
2- Yes, it's on it's way
3- I have seen that, The tests failed when I used `tempfile` and I wanted to investigate the problem so I ran them with the hardcoded path. Will change that.

warenlg:
Awesome @glimow !

Agreed with Konst. Also I think it could be nice to squash your similar commits like "fix checks" for example so that we keep a relevant commit history.

glimow:
> Awesome @glimow !
> 
> Agreed with Konst. Also I think it could be nice to squash your similar commits like "fix checks" for example so that we keep a relevant commit history.

You are right. Commits Squashed. Highlights in @zurk's comment taken into account as well. PR is almost finished, last question is where to get the model weights from when using it in `TokenParser`

zurk:
>  last question is where to get the model weights from

I think they should be included in the modelforge model. You should load the model and take them from it. Do I miss something? 

Also, tests are not passed. PTAL
https://travis-ci.com/src-d/ml/jobs/193718842#L1680-L1743

glimow:
Yes, tests do not pass due to tf misconfig. They worked before so I have to find the commit where things got mixed up. I will ask @irinakhismatullina to help me uploading the weights when they will pass

zurk:
:+1: 
`git bisect` is your best friend in such a case.

glimow:
NOTE: Tests pass, but travis is laggy !

sourced/ml/models/id_splitter.py
```diff
@@ -0,0 +1,124 @@
...
+from keras.preprocessing.sequence import pad_sequences
+from modelforge import Model, register_model
+import numpy as np
+# import tensorflow as tf
+
+from sourced.ml.algorithms.id_splitter.nn_model import (f1score, precision,
+                                                        recall)
+from sourced.ml.models.license import DEFAULT_LICENSE
+
+MAXLEN = 40
```
zurk: I think `MAXLEN` is model dependent and we should include it to the model.

sourced/ml/models/id_splitter.py
```diff
@@ -0,0 +1,124 @@
...
+
+        return processed_ids, clean_ids
+
+    def load_model_file(self, path="./id_splitter_rnn.model"):
+
+        self._model = keras.models.load_model(path, custom_objects={"precision": precision,
+                                                                    "recall": recall,
+                                                                    "f1score": f1score})
+
+    def __call__(self, tokens: [str]) -> [[str]]:
```
zurk: let's call it `split()`

zurk:
```suggestion
    def __call__(self, tokens: Sequence[str]) -> List[str]:
```
please refer to https://docs.python.org/3/library/typing.html

sourced/ml/algorithms/token_parser.py
```diff
@@ -33,14 +33,16 @@ class TokenParser:
     # if True we have only ["sourc", "algorithm"].
     # if you do not want to filter small tokens set min_split_length=1.
     SAVE_TOKEN_STYLE = False  # whether yield metadata that can be used to reconstruct initial
-    # identifier.
+
+    # identifier
+    USE_NN = False  # wether to use neural network id splitter model instead of heuristic
```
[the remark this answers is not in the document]
glimow: waiting for @vmarkovtsev's opinion. I agree with you on that but I discussed that with sb and finished by adding that parameter to the original class. Can't remember the exact circumstances.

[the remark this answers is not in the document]
glimow: If I understood well @vmarkovtsev we stick with one `TokenParser` class for now ?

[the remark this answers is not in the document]
zurk: yes

sourced/ml/models/id_splitter.py
glimow: Included in the model, this is just the default value.

irinakhismatullina: Still it's better to put it inside the class, like in `TokenParser`. We do it like that everywhere, if it's not shared among several classes or functions.

sourced/ml/models/id_splitter.py
```diff
@@ -0,0 +1,125 @@
...
+        keras.backend.clear_session()
+
+    def _generate_tree(self) -> dict:
+        return {
+            "config": self._model.get_config(),
+            "weights": self._model.get_weights()
+            }
+
+    def _load_tree(self, tree: dict):
+
```
irinakhismatullina: Why the newlines in the beginning of the functions? Here and in some places lower.

sourced/ml/models/id_splitter.py
```diff
@@ -0,0 +1,125 @@
...
+            clean_id = clean_id[:MAXLEN]
+        logging.info("Preprocessed identifier: {}".format(clean_id))
+        return np.array([mapping[c] for c in clean_id], dtype="int8"), clean_id
+
+    def prepare_input(self, identifiers: Sequence[str], maxlen: int = MAXLEN,
+                      padding: int = PADDING, mapping: dict=None) -> Tuple["np.array", List[str]]:
+        """Prepares input by converting a sequence of identifiers to the corresponding
+        ascii code 2D-array and the list of lowercase cleaned identifiers.
+        """
+
```
irinakhismatullina: Here.

sourced/ml/models/id_splitter.py
```diff
@@ -0,0 +1,125 @@
...
+        clean_ids = []
+        for identifier in identifiers:
+            feat, clean_id = self._prepare_single_identifier(
+                identifier, maxlen=maxlen, padding=padding)
+            processed_ids.append(feat)
+            clean_ids.append(clean_id)
+
+        processed_ids = pad_sequences(processed_ids, maxlen=maxlen,
+                                      padding=padding)
+
```
irinakhismatullina: Actually it's better to remove some unnecessary newlines, like here - it doesn't affect readability.

sourced/ml/models/id_splitter.py
```diff
@@ -0,0 +1,125 @@
...
+        processed_ids = []
+        clean_ids = []
+        for identifier in identifiers:
+            feat, clean_id = self._prepare_single_identifier(
+                identifier, maxlen=maxlen, padding=padding)
+            processed_ids.append(feat)
+            clean_ids.append(clean_id)
+
+        processed_ids = pad_sequences(processed_ids, maxlen=maxlen,
```
irinakhismatullina: Here fits in one line

sourced/ml/tests/test_id_splitter_nn_model.py
```diff
@@ -1,12 +1,15 @@
...
 import unittest
 
 import numpy
 
 from sourced.ml.tests import has_tensorflow
+from sourced.ml.tests.models import ID_SPLITTER_RNN
 
 
 class MetricsTests(unittest.TestCase):
+
```
irinakhismatullina: Newline

sourced/ml/tests/test_id_splitter_nn_model.py
```diff
@@ -52,5 +55,27 @@ def test_build_cnn(self):
         self.assertTrue(self.model_cnn.uses_learning_phase)
 
 
+class NNModelTest(unittest.TestCase):
+
```
irinakhismatullina: Here too

sourced/ml/algorithms/token_parser.py
```diff
@@ -179,6 +196,7 @@ def reconstruct(tokens):
             elif meta == TokenStyle.TOKEN_CAPITALIZED:
                 res.append(t[0].upper() + t[1:])
         return "".join(res)
+    # def _nn_split(self, tokens):
```
irinakhismatullina: Please remove commented unused code

sourced/ml/models/id_splitter.py
```diff
@@ -0,0 +1,122 @@
...
+            "config": self._model.get_config(),
+            "weights": self._model.get_weights()
+            }
+
+    def _load_tree(self, tree: dict):
+        model = keras.models.Model.from_config(tree["config"])
+        model.set_weights(tree["weights"])
+        self.construct(model=model)
+
+    def _prepare_single_identifier(self, identifier: str, maxlen: int=MAXLEN, padding: int=PADDING,
```
irinakhismatullina:
```suggestion
    def _prepare_single_identifier(self, identifier: str, maxlen: int = MAXLEN, padding: int = PADDING,
```
And the same everywhere lower

sourced/ml/models/id_splitter.py
```diff
@@ -0,0 +1,122 @@
...
+                                   mapping: dict=None) -> "np.array":
```
zurk:
```suggestion
                                   mapping: dict=None) -> np.array:
```
Here and in many places, it should work without quotes.
Usually, you need quotes when the type of return value is unavailable e.g. `IdentifierSplitterNN.construct()` returns  `IdentifierSplitterNN` and it is not ready yet when we interpret `construct` function defenition

sourced/ml/models/id_splitter.py
irinakhismatullina: Newlines?

sourced/ml/models/id_splitter.py
```diff
@@ -0,0 +1,122 @@
...
+        if mapping is None:
+            mapping = {c: i for i, c in enumerate(string.ascii_lowercase, start=1)}
+
+        # Clean identifier
+        clean_id = "".join([char for char in identifier.lower() if char in string.ascii_lowercase])
```
zurk:
```suggestion
        clean_id = "".join(char for char in identifier.lower() if char in string.ascii_lowercase)
```

sourced/ml/models/id_splitter.py
```diff
@@ -0,0 +1,122 @@
...
+        if len(clean_id) > MAXLEN:
```
zurk: you have all the parameters inside the class, let's use them:
`self._maxlen` instead of `MAXLEN`. `maxlen` should be removed from the function parameters. 

`padding` should be removed also.

sourced/ml/models/id_splitter.py
zurk: And the return type is wrong should be `Tuple[numpy.array, str]`

sourced/ml/models/id_splitter.py
```diff
@@ -0,0 +1,122 @@
...
+            clean_id = clean_id[:MAXLEN]
+        logging.info("Preprocessed identifier: {}".format(clean_id))
+        return np.array([mapping[c] for c in clean_id], dtype="int8"), clean_id
+
+    def prepare_input(self, identifiers: Sequence[str], maxlen: int = MAXLEN,
+                      padding: int = PADDING, mapping: dict=None) -> Tuple["np.array", List[str]]:
```
zurk: same here. You do not need `maxlen` and `padding`, they are in class already.

sourced/ml/models/id_splitter.py
```diff
@@ -0,0 +1,122 @@
...
+        """Prepares input by converting a sequence of identifiers to the corresponding
```
zurk:
```suggestion
        """
        Prepares input by converting a sequence of identifiers to the corresponding
```

sourced/ml/models/id_splitter.py
```diff
@@ -0,0 +1,122 @@
...
+        ascii code 2D-array and the list of lowercase cleaned identifiers.
+        """
```
zurk: please add `mapping` param and return value description to docstring. 
Should it be included to model itself?

zurk: And I do not see `mapping` usage in the function...

sourced/ml/models/id_splitter.py
```diff
@@ -0,0 +1,122 @@
...
+            feat, clean_id = self._prepare_single_identifier(
+                identifier, maxlen=maxlen, padding=padding)
+            processed_ids.append(feat)
+            clean_ids.append(clean_id)
+
+        processed_ids = pad_sequences(processed_ids, maxlen=maxlen, padding=padding)
+
+        return processed_ids, clean_ids
+
+    def load_model_file(self, path=None):
```
zurk:
```suggestion
    def load_model_file(self, path: str):
```

sourced/ml/models/id_splitter.py
```diff
@@ -0,0 +1,122 @@
...
+        assert path is not None
```
zurk:
```suggestion
```

sourced/ml/models/id_splitter.py
zurk: And I do not see function usage. delete it?

sourced/ml/models/id_splitter.py
```diff
@@ -0,0 +1,122 @@
...
+        self._model = keras.models.load_model(path, custom_objects={"precision": precision,
+                                                                    "recall": recall,
+                                                                    "f1score": f1score})
+
+    def split(self, tokens: Sequence[str]) -> List[str]:
+        """
+        Splits a lists of tokens using the model.
+        """
```
zurk: ofc, it is kind of obvious, but let's add `tokens` and return value description. and I think by `tokens` you mean `identifiers`, let's rename it if so.

sourced/ml/tests/test_id_splitter_nn_model.py
```diff
@@ -52,5 +54,26 @@ def test_build_cnn(self):
...
+        self.test_y = ["network", "socket", "variable",
+                       "name", "load", "file", "blah", "blah", "foobar"]
+        self.id_splitter = IdentifierSplitterNN()
+        self.id_splitter.load(ID_SPLITTER_RNN)
+
+    @unittest.skipIf(not has_tensorflow(), "Tensorflow is not installed.")
+    def test_load_and_run_model(self):
+        self.assertEqual(self.id_splitter.split(self.test_X), self.test_y)
+
+    @unittest.skipIf(not has_tensorflow(), "Tensorflow is not installed.")
```
zurk: you can use a decorator for all the `NNModelTest` class

sourced/ml/models/id_splitter.py
```diff
@@ -0,0 +1,122 @@
+import logging
+import os
+import string
+from typing import List, Sequence, Tuple
+
+import keras
+from keras import backend
+from keras.preprocessing.sequence import pad_sequences
+from modelforge import Model, register_model
+import numpy as np
```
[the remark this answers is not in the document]
glimow: Personally I don't care, but `import numpy as np` is used in the whole official numpy documentation

sourced/ml/models/id_splitter.py
glimow: ok

sourced/ml/models/id_splitter.py
glimow: ok. I worked in functional programming for too long lol.

sourced/ml/models/id_splitter.py
glimow: This function can be used to convert keras models into `modelforge` models by loading the keras model file and saving the model using model.save() method.

sourced/ml/models/id_splitter.py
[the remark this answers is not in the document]
zurk: yes, but If you check our code we do not use it at all.
the same for pandas, scikit-learn, etc.

sourced/ml/models/id_splitter.py
zurk: ok, then docstring it, please.

sourced/ml/models/id_splitter.py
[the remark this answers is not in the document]
glimow: ok

sourced/ml/models/id_splitter.py
glimow: ok

sourced/ml/models/id_splitter.py
glimow: ok

sourced/ml/algorithms/token_parser.py
```diff
@@ -166,6 +169,20 @@ def ret(name):
             if last:
                 yield from ret(last)
 
+    def split(self, token):
+        """Splits a single identifier
```
zurk: we do not use such style.
```
        """
        Splits a single identifier
        """
```
or 
```
        """Splits a single identifier"""
```

And please add `.` in the end.

Please check all docstrings

sourced/ml/algorithms/token_parser.py
```diff
@@ -166,6 +169,20 @@ def ret(name):
...
+        """
+        if self._use_nn:
+            return self.id_splitter_model([token]).pop()
+        return self._split(token)
+
+    def split_batch(self, tokens):
+        """Splits a batch of identifiers
```
zurk: same here

sourced/ml/models/id_splitter.py
```diff
@@ -0,0 +1,124 @@
...
+    DEFAULT_MAXLEN = 40
+    DEFAULT_PADDING = "post"
+    DEFAULT_MAPPING = None
+
+    def construct(self, model: "keras.models.Model" = None,
+                  session: "tf.Session" = None,
+                  maxlen: int = DEFAULT_MAXLEN,
+                  padding: str = DEFAULT_PADDING,
+                  mapping: dict = DEFAULT_MAPPING) -> "IdentifierSplitterNN":
+        assert model is not None
```
zurk: if you need `model` just do not set default argument value.

sourced/ml/models/id_splitter.py
```diff
@@ -0,0 +1,124 @@
...
+
+        if session is not None:
+            backend.tensorflow_backend.set_session(session)
+        os.environ["TF_CPP_MIN_LOG_LEVEL"] = "1"
```
zurk: I do not think we should set it here because it is unobvious for the user. If it is required, the user can set it by himself

sourced/ml/models/id_splitter.py
```diff
@@ -0,0 +1,124 @@
...
+            "config": self._model.get_config(),
+            "weights": self._model.get_weights()
+            }
+
+    def _load_tree(self, tree: dict):
+        model = keras.models.Model.from_config(tree["config"])
+        model.set_weights(tree["weights"])
+        self.construct(model=model)
+
+    def _prepare_single_identifier(self, identifier: str) -> numpy.array:
```
zurk: return type is wrong

sourced/ml/models/id_splitter.py
```diff
@@ -0,0 +1,124 @@
...
+        if self._mapping is None:
+            self._mapping = {c: i for i, c in enumerate(string.ascii_lowercase, start=1)}
```
zurk: ok, I think I get what kind of mapping is it.
It looks like mapping should be the part of the model itself because who knows how I map the characters during training.

sourced/ml/models/id_splitter.py
```diff
@@ -0,0 +1,124 @@
...
+
+        # Clean identifier
+        clean_id = "".join(char for char in identifier.lower() if char in string.ascii_lowercase)
```
zurk:
```suggestion
        clean_id = "".join(char for char in identifier.lower() if char in self._mapping)
```

sourced/ml/algorithms/token_parser.py
zurk: Can you also add another test to check TokenParser with use_nn enabled?

glimow: I have a call with my university soon, no prob I'll do that just afterwards.
I should have fixed the rest, let's see if the test passes.

